PR #2, opened 2020-04-26
Init cube section plot docs

amoodie:
This PR represents a significant amount of development on the DeltaMetrics framework.

## before merging
* I'm principally looking for feedback on how the API works / feels.
* I did a lot of documentation and some testing, but both need to be expanded.

You can see a demonstration of how I set up the API in [the 10 minute tutorial](https://raw.githubusercontent.com/amoodie/DeltaMetrics/be4520470ed192ef8ee560deb484a91ca65ff2bf/docs/source/guides/10min.rst). You may want to checkout this branch and build the docs for yourself locally to see it nicely rendered.

## changes
* a class that handles the file reading. `io.py` defines a base class that other file formats inherit from. This is to enforce a consistent API for all file i/o formats. I expect that this class and API will change when we settle on a [format and structure for files in pyDeltaRCM](https://github.com/DeltaRCM/pyDeltaRCM_WMT/issues/13). Does not load anything into memory by default.
* a class that defines the Cube, and handles slicing, accessing data, some initial plotting routines. Can access anything inside the underlying data file without loading into memory, but can load into memory for faster operations (*not fully implemented or tested*). Computes stratigraphy *mapping* (see not below); computing the index rather than placing into a new array allows any variable to be transformed to stratigraphy via the same t-x-y to z-x-y mapping.
* a class that defines the Section, which handles variable access, some plotting, etc. Most methods are defined in BaseSection, and the subclasses need to compute the x-y coordinate pairs to extract for the section (I still need to fix the implementation of this, and compute distance between extract points for the along-section coordinate system `s`).
* a class that defines some helpful plotting utilities. This helps make all the plots use consistent color schemes for each variable. This utility can link colormaps across cube instances, and can be fully customized. (*need to do tests and documentation on all this still too*).

## a note on stratigraphy
I have implemented a method for computing stratigraphy that does not convert to a regular z-x-y grid (like the [`strat_preprocess.py`](https://github.com/DeltaRCM/pyDeltaRCM_WMT/blob/develop/strat_preprocess.py) file in pyDeltaRCM does. Instead, I just use a mesh defining the `eta` and `s` (`s` is the along-section coordinate) of the section of interest and then plot that. I think we should also implement a method for "boxy" stratigraphy.


docs/source/meta/installing.rst
```diff
@@ -2,12 +2,44 @@
...
+    $ git clone https://github.com/DeltaRCM/DeltaMetrics.git
 
-.. note::
-	
-	THE PACKAGE IS NOT AVAILABLE ON PYPI YET
+and then ``cd`` into the directory and install editable copy.
+
+.. code:: console
+
+    $ cd deltametrics
```
elbeejay: Nit-picking, but I would prefer `cd DeltaMetrics` -- at least that's how the directory is created for me when I clone it.

amoodie: yep, I got it right in the dev guide, but wrong here. Fixed, and will commit.

docs/source/meta/planning.rst
```diff
@@ -12,7 +12,7 @@ The goal of the DeltaMetrics project is to provide convenient tools to analyze s
...
+#. Users should be able to instantiate the Cube object with a numpy array where dimensions are expected to be ``t-x-y``, and a value specifying the type of data stored in the array. Data could be, lidar scans, overhead photos, grain size maps (DeltaRCM), flow velocity (DeltaRCM), etc. An optional time array can be supplied, specifying time at each z; we can also accept a single value ``T``, and time is linearly spaced wrt ``T`` and ``0``.
 #. Users should be able to extract ``Section`` s (e.g., :class:`~deltametrics.section.RadialSection`) from the ``Cube``. Several types of sectioning should be supported. By default, we should try to process to "stratigraphy" via an elevation timeseries attribute of the ``Cube``, but can have a flag to leave as raw section.
 #. Users should be able to do processed and compute statistics on the ``Section`` s. For example, idenfying channels (?) or computing compensation statistics (extract a ``Horizon`` first?). These routines can live in ``deltametrics.strat`` (:doc:`../reference/strat/index`) or in other modules.
```
elbeejay, line 17: I like the framework of the `Cube` from which either Sections or Plans are extracted and metrics are computed. In this format, where would methods for 3-D or bulk stratigraphic metrics be computed (e.g. net-gross)? It makes me wonder if the `Section` should be named `Stratigraphy` or something else? Unless those 3-D stratigraphic metrics are just going to be housed in `Cube` (which is fine by me).

docs/source/pyplots/section/section_lexicon.py
```diff
@@ -0,0 +1,13 @@
+import matplotlib.pyplot as plt
+
+import deltametrics as dm
+
+rcm8cube = dm.sample_data.cube.rcm8()
+rcm8cube.register_section('demo', dm.section.StrikeSection(y=10))
+
```
elbeejay, line 8: This is a nice example. Functionality-wise, I like the loading of the cube and then the slicing of the section.

When the `rcm8cube` object is created it is basically loading (via the `io` wrapper etc) the NetCDF4 file. So the *variables* available can be listed using `rcm8cube.variables`. However once we move to the section `rcm8cube.sections`, we lose the ability to read the variables available in the section. Is the list of variables something worth passing along to the section? Unless you know to check `rcm8cube.variables`, you don't know that in addition to 'strata_sand_frac', you could call 'velocity' when visualizing the section.
 
A lot of the I/O and variable handing seems to have been done carefully with respect to name-spaces and memory, but my ham-handed suggestion is something like `self[name].variables = self.variables` in the `register_section` function

deltametrics/cube.py
```diff
@@ -1,21 +1,406 @@
...
 
-        pass
+        # self.data = data
+        # self.data = _DataDict({})
+        if type(data) is str:
+            self._data_path = data
+            self._connect_to_file(data_path=data)
+            self._read_meta_from_file()
+            # self._connect_vars_to_file(self.variables)
+            self._compute_strata()
```
ericbarefoot: How come this is part of the `__init__`? It seems to me like this computation should be an 'if requested' feature. especially if it can be slow as you mention for large datasets.

deltametrics/cube.py
```diff
@@ -1,21 +1,406 @@
...
+        elif type(data) is dict:
+            # handle a dict, arrays set up already, make an io class to wrap it
+            self._data_path = None
+            raise NotImplementedError
+        else:
+            raise TypeError('Invalid type for "data": %s' % type(data))
```
ericbarefoot, line 137: This is a good idea. This reminds me that there should be, along with the rich set of features we're doing here, some way for people to use really simple datasets, and we just make strong assumptions about the data. Not an issue here, but just a reminder for us later.

deltametrics/cube.py
```diff
@@ -1,21 +1,406 @@
...
+        -------
+        CubeVariable : `~deltametrics.cube.CubeVariable`
+            The instantiated CubeVariable.
+        """
+
+        if var in self._variables:
+            return CubeVariable(self.dataio[var], variable=var)
+        elif var == 'time':
+            # a special attribute we add, which matches eta.shape
+            _eta = self.dataio[var]
```
ericbarefoot, line 168: Is `_eta` intended to be just limited to actual elevation variable itself? I am curious because I see eta elsewhere as a special case, rather than var. But other things can vary with time too, right?

deltametrics/cube.py
```diff
@@ -1,21 +1,406 @@
...
+        """Connect to file.
+
+        This method is used internally to send the ``data_path`` to the
+        correct IO handler.
+        """
+        _, ext = os.path.splitext(data_path)
+        if ext == '.nc':
+            self._dataio = io.NetCDFIO(data_path)
+        elif ext == '.hf5':  # ?????
+            self._dataio = io.HDFIO(data_path)
```
ericbarefoot, line 186: Yah. I can get on this.

deltametrics/cube.py
```diff
@@ -1,21 +1,406 @@
...
+
+        Currently, we store many of these computations as private attributes
+        of the Cube itself. This is fine for a small cube, but may become
+        cumbersome or intractable for large datasets.
+        problematic as the matrix becomes very large.
+
+        .. note::
+            This could be ported out to wrap around a generalized function for
+            computing stratigraphic representations of cubes. E.g., wrapping
+            some function in :obj:`~deltametrics.strat`.
```
ericbarefoot: I think this is probably a good idea, so that the user can request the computation as a request. What if they've already done a preservation calculation?

deltametrics/sample_data/cube.py
```diff
@@ -18,19 +23,32 @@
...
 import numpy as np
+import netCDF4
+
+from ..cube import Cube
+from ..io import NetCDFIO, HDFIO
+
 
 def tdb12():
-  return np.array([[]])
+    raise NotImplementedError
```
ericbarefoot, line 44: The whole thing would be kind of a pain since just the topo is about 2 GB, but we could provide a small T x X x Y square cube.

deltametrics/section.py
```diff
@@ -1,10 +1,180 @@
...
+    def as_strata(self):
+        return self._strata
+
+
+class SectionVariable(object):
+    """Section variable.
+
+        This could be one way to implement functionality of section
+        variables. We would have to implement all the functionality of the
+        ndarray ourselves. It's more readable to do it this way, but much
```
ericbarefoot: Why would we have to implement everything from `ndarray` ourselves? This isn't clear to me.

deltametrics/io.py
```diff
@@ -0,0 +1,217 @@
...
+    def write(self):
+        """Write data to file.
+
+        Take a :obj:`~deltametrics.cube.Cube` and write it to file.
+
+        .. warning::
+            Not Implemented.
+
+        """
+        raise NotImplementedError
```
ericbarefoot, line 188: Woof. This is going to be a hard one to do, just because there's all kinds of contingencies to  plan for, including how people want to compress and write to disk.

deltametrics/io.py
```diff
@@ -0,0 +1,217 @@
...
+        Parameters
+        ----------
+        var : `str`
+            Which variable to load from the file.
+        """
+        try:
+            _arr = self.dataset.variables[var]
+        except ValueError as e:
+            raise e
+        self._in_memory_data[var] = np.array(_arr, copy=True)
```
ericbarefoot, line 177: Is there a way to warn the person if they're about to load a really large dataset into memory? Say, it the array is going to be >4 GB of ram, just raise a warning to make sure they know?

ericbarefoot (approved):
In general, I like the way this works. I am still a little unclear on some aspects of it, like how the user will access the meta-data and link across variables. In general, the data I/O looks good, I think the writing to file and supplying the arguments is going to be challenge.

I'm intrigued by the notion of having a preserved voxel arrays, but I'm a little unsure about it being computed on-the-fly every time the cube is instantiated. Also, let's say someone has an array they want to use as a cube, but they've already done their own method for computing the synthetic stratigraphy. I like the idea of having a lot of convenience methods for stuff that lots of people want to do, but if possible, my preference would be to have the base `Cube` be a little more data content agnostic. For example, not assuming that there is any variable called `eta` at all.

I like the plotting stuff is really clever. I can't get anything to show up on mine, but also I think my laptop and matplotlib just don't play nice to each other. For this reason, I might advocate as a future thing putting in a way to send plots to disk as files.

elbeejay (commented):
Thanks for getting this all organized and structured Andrew, it is a big step forward to making this something that's relatively easy to work on and contribute to. 

The overarching `Cube`, `Plan`, and `Section` framework makes it clean which sections of the code do what and seems like a good format to keep things. The thing I am struggling with (in-part because I don't have a better suggestion) is how the different functionalities of the code tie together. If you have instantiated a `Cube` and you want to register a section, you have to both `.register_section` and then call upon the `section.StrikeSection` function. Could this be streamlined by declaring the type of section you wish to register rather than calling another function, like `.register_section(name='example',type='strike',location=5)` or is there a larger flexibility/readability component I am missing here?

I get the impression that the `plan` and `mask` sets of objects are similarly related, which begins to create a real tree of function calls across different files when you start with `register_plan` and build from there. I acknowledge that we are stuck with some sort of system for handling multiple files and functions if we want a multi-functional code, just wanted to point this out. I guess hard-impositions or assumptions about the type of `mask` a `plan` function will use starts to limit the user's flexibility when using the code and limits us to a single method of masking.

All in all, I think I can get behind this general structure for the codebase. It is quite clear as far as where certain functions ought to be built and the format they must be built in to work with the rest of the code.

elbeejay (approved):

docs/source/meta/planning.rst
amoodie: I *think* what we should do, is have the modules `strat.py` and `channel_network.py` or `overland_flow.py` or whatever that have all the fancy algorithms, computations, statistics, etc. Leave `cube.py` `section.py` and `plan.py` to set up the objects that these algorithms take as arguments. We could "wrap" a few of the most commonly used ones with methods of the object classes too, if needed. 

I mentioned somewhere in a docstring about porting the strat calculations out to a module. This may be the right move, but we can always do it in the future.

docs/source/pyplots/section/section_lexicon.py
amoodie: yep, definitely a good idea. I'll implement something like this.

deltametrics/cube.py
amoodie: Okay, I'm fine moving this out of `__init__`. I'll make this change. 

I'm going to add an optional argument to compute stratigraphy on instantiation though: `strata_from='eta'`, would compute strata during `__init__` from the `eta` variable.

deltametrics/cube.py
amoodie: I'm not really sure what you mean, but yeah. I tried to set up the structure of the package to be expandable and flexible.

deltametrics/cube.py
amoodie: `_eta` is just a view of the `eta` variable, created temporarily, to access it's shape attribute. I guess we could get the array shape during `Cube` instantiation, and then store it. That may be better.

Ultimately though, this is likely to have to change anyway when we create a new standard for the file structures.

Yeah of course, anything can vary with time. Right now, I have things set up to assume that all variables have a common "time" vector. As we've discussed, we may want to have time be a vector defined for each variable individually. If that's the case, we would create the time

> I see eta elsewhere as a special case, rather than var.

I guess this is probably because `eta` dictates what gets preserved in stratigraphy, whereas `sand_frac`, `velocity`, etc, do not. So `eta` is kind of special in that way.

deltametrics/cube.py
amoodie: Not critical right now, I can write the IO for DeltaMetrics too, I don't mind.

What we need to do though, is decide on a standard format for storing the data on disc. Some well-defined, self describing dataset that we can work with. 

It's hard to write this DeltaMetrics package without that standard "expected" way of getting data. -------In the distant future, we can expand the IO to handle all sorts of other things.

deltametrics/sample_data/cube.py
amoodie: yeah that's fine. Not worried about it for now. Just put there as kind of a placeholder/reminder.

deltametrics/section.py
amoodie: If we create a new object, it doesn't have a, e.g., `__add__` method by default. 

Alternatively, we can *subclass* the ndarray, and then we get all of its methods. I want to do this, but I haven't quite figured it out.

deltametrics/io.py
amoodie: hahah that's fine, I'm not worried about leaving it not implemented for now. We can worry about it in the future. 

Ultimately, I think we should *think* about making the package flexible for lots of use cases, but primarily we want to make the parts that we need first, and worry about other stuff later.

deltametrics/io.py
amoodie: Sure, I can throw a warning in there. Would you like it to be a "reject" warning, where we require a `force=True` to override the warning and load anyway?

The reading is disabled by default though, so if someone tries to read a 4GB file...well, swim at your own risk...

amoodie:
@elbeejay 

> If you have instantiated a Cube and you want to register a section, you have to both .register_section and then call upon the section.StrikeSection function. Could this be streamlined by declaring the type of section you wish to register rather than calling another function, like .register_section(name='example',type='strike',location=5) or is there a larger flexibility/readability component I am missing here?

Yep, that's :100: something we could do in. For now, I think I'll just leave it as is. Once we settle on the keywords used to create the sections of all the different types, then we can pretty easily extend the `register_section` function to take either 1) a valid `Section` instance of some type (current implementation) or 2) a set of kwargs that allow a valid section to be instantiated.


> I get the impression that the plan and mask sets of objects are similarly related, which begins to create a real tree of function calls across different files when you start with register_plan and build from there. 

I'm absolutely not sure what module names, and functionalities we're going to wind up with. We're in early development now, we can change the API 100% any day we want. I just tried to set things up how I think they may work. If, when you implement some functions, want to set up the modules differently, I'm all for it.

In general though, like I mentioned elsewhere, I *think( what we should do is have the modules `strat.py` and `channel_network.py` or `overland_flow.py` or whatever that have all the fancy algorithms, computations, statistics, etc. Leave `cube.py` `section.py` and `plan.py` to set up the objects that these algorithms take as arguments. We could "wrap" a few of the most commonly used ones with methods of the object classes too, if needed.

amoodie:
@ericbarefoot 

> I am still a little unclear on some aspects of it, like how the user will access the meta-data and link across variables.

I think that the `Cube` should expose the relevant metadata with convenient tags, e.g., `cube.variables`. Beyond that, we can add all sorts of methods, and the documentation will allow users to find in the `IO` objects, how to list metadata out. Our `io.py` just wraps around the other packages, so it should be easy for anyone to explore the file on disk just as they would with the regular module.

> In general, the data I/O looks good, I think the writing to file and supplying the arguments is going to be challenge.

Yeah, probably. We can ignore it for now, and decide to implement it later. If we do implement it, we just have to worry about doing it how we want to write the files. If someone wants to do it differently, the can write their own write functions!

> I'm intrigued by the notion of having a preserved voxel arrays, but I'm a little unsure about it being computed on-the-fly every time the cube is instantiated. 

Yeah, I mean, in theory, you would only instantiate a cube once and then use it throughout your session. But regardless, I'll move the computation out of `__init__` and make it an optional task.


> Also, let's say someone has an array they want to use as a cube, but they've already done their own method for computing the synthetic stratigraphy. I like the idea of having a lot of convenience methods for stuff that lots of people want to do, but if possible, my preference would be to have the base Cube be a little more data content agnostic. For example, not assuming that there is any variable called eta at all.

Yeah, okay. What about this. 
* We have `BaseCube` which is super data agnostic, just handles the slicing setup, variable names, IO, etc. The user will not interact with this object directly. 
* Then we can make `DataCube` which inherits from base, and would be the type of cube in the example file; This object has the ability to compute stratigraphy based on a variable name (default would be `eta`, but could be anything). The `Cube` I created is mostly this one.
* Next we could have `StratigraphyCube` which is hard-coded stratigraphy, only having knowledge of preserved data; in your example, the user would instantiate this object.
* we could have ImageCube that is expected to be a series of RGB images of the delta (supporting experiments, or satellite?)
* etc as needed

Thoughts on this???! @elbeejay 

> For this reason, I might advocate as a future thing putting in a way to send plots to disk as files.

Not sure what you mean...In the `10min` example, you would just `plt.savefig('path/to/file.png')` instead of `plt.show()`. Or do you mean like saving images without matplotlib?

elbeejay:
> Yeah, okay. What about this.
> 
> - We have `BaseCube` which is super data agnostic, just handles the slicing setup, variable names, IO, etc. The user will not interact with this object directly.

> - Then we can make `DataCube` which inherits from base, and would be the type of cube in the example file; This object has the ability to compute stratigraphy based on a variable name (default would be eta, but could be anything). The Cube I created is mostly this one.

I support this, I very much do like the current structure of the `Cube`. Changing the name to be a bit more descriptive and allow for other similar types of cubes to exist is not an issue with me. 

> - Next we could have `StratigraphyCube` which is hard-coded stratigraphy, only having knowledge of preserved data; in your example, the user would instantiate this object.
> - we could have `ImageCube` that is expected to be a series of RGB images of the delta (supporting experiments, or satellite?)
> - etc as needed

Sure, if breaking it out a little bit to gain this flexibility (for experimental strata, satellite images) is what makes the most sense I don't object. Might give us a bit more ability to write in default or canned routines for some of these "cubes" if they are a bit more narrowly defined than a very general single datacube.

deltametrics/io.py
ericbarefoot: Lol. Yeah I agree that people shouldn't do it unless they know what they're doing, I foresee this being a thing that someone might do while exploring the API, and just try to read all variables from a 50 GB dataset into memory, without realizing it. I think a `force=True` warning would be a great way to put this in there, potentially with a short call to `psutil` at the beginning to query the system and figure out what the total memory capacity is, and only throw a warning if the file is, say >80% of the total available.

deltametrics/io.py
ericbarefoot: I agree, let's just leave this unimplemented for now.

deltametrics/section.py
ericbarefoot: Ok, I think I just don't understand the non-inheritance here. Don't worry about it, I'll learn more later, and as it stands, I think this looks good!

deltametrics/sample_data/cube.py
ericbarefoot: I'll send something to you and Jay to play with.

deltametrics/cube.py
ericbarefoot: >I'm going to add an optional argument to compute stratigraphy on instantiation though: `strata_from='eta'`, would compute strata during `__init__` from the eta variable.

That's a good workaround. I agree this is such a common thing, that it's good to have it as part of the loading process, I wonder if you could save this sparse matrix as part of the on-disk file too?

deltametrics/cube.py
ericbarefoot: Like we talked about in Zoom, but I will just put here for completeness. I was imagining the use-case where someone just has a pre-made cube of data, and they just want to visualize slices and calculate compensation or whatever, there should be some simple interface for them. I wasn't clear about how this dict framework would work, but I understand better now.

deltametrics/cube.py
amoodie: yeah, if the need arises, we can definitely provide the option to write the strat array info to file to load it again next time. Skipping this for now though.

deltametrics/cube.py
amoodie: I created a `StratigraphyCube` that will allow for these things, and a Section type to go with it. It's not really implemented now though, but the placeholders and setup is there, I think. Need some sample data to really try to make it work in the future.

deltametrics/section.py
amoodie: I am now subclassing the MaskedArray for all of the `SectionVariable` types, so they will all support arbitrary math and slicing :+1:

deltametrics/cube.py
amoodie: This isn't exactly ported out to `strat` currently, but it has been logically separated from initialization, which I think satisfies the point from an api perspective.

amoodie:
@ericbarefoot @elbeejay lmk if you want to have a look at the latest changes, otherwise, I implemented most of what we talked about.

Overall, I think a fair amount of the stratigraphy and section stuff will change as we figure our how we want to use and work with the data and objects. I think it's good for now, but just wanted to be clear I'm not married to anything here, and we can change anything we want at any time.

PR #7, opened 2020-06-28
Masking

elbeejay:
### Overall
~~Don't think this is quite ready to be merged yet, but~~ I am looking for some feedback on the structure of the masking classes/functions. I am a bit concerned about flexibility and ease of implementation for other methods with the structure I've implemented here.

#### Changes Proposed
- Automated shoreline detection by way of the opening angle method as a `ShoreMask` class
- Land mask creation using by-products of the shoreline detection as a `LandMask` class
- Surface water or wet-mask as a `WetMask` class
- Some tests for the mask functionality were added and documentation was updated as necessary

#### To-Do
Before merging, I'm planning to add the following masking functions:
- [x] A channel mask
- [x] A centerline mask
- [x] An "edge" mask (all water/land interfaces) 
- [x] Tests for the channel mask and 'edge' mask
- [x] Verify that `rivamap` functionality works (at least locally)

Happy to try to add other functionality while digging into this mask stuff, so let me know if there are any features we have been wanting to add to the `mask.py` script.


deltametrics/mask.py
```diff
@@ -48,51 +53,308 @@ def show(self, **kwargs):
         """
         cmap = kwargs.pop('cmap', 'gray')
         fig, ax = plt.subplots()
-        ax.imshow(self.data, cmap=cmap, **kwargs)
+        if hasattr(self, 'mask'):
+            ax.imshow(self.mask, cmap=cmap, **kwargs)
+            ax.set_title('A ' + self.mask_type + ' mask')
+        else:
```
amoodie, line 61: I think I would prefer to just throw an error here, rather than plotting the input. My opinion is that it is harder to debug when you don't get what you want, but also don't get an error.

deltametrics/mask.py
```diff
@@ -1,6 +1,11 @@
...
-from scipy import stats
 import matplotlib.pyplot as plt
+from scipy.spatial import ConvexHull
+from shapely.geometry import Point
+from shapely.geometry.polygon import Polygon
+from skimage import feature
+from skimage import morphology
 
 
 class BaseMask(object):
```
amoodie, line 12: add to the `__init__` (see later comments, for this to make sense):

```
self._mask = np.zeros(self.data.shape)
```

deltametrics/mask.py
```diff
@@ -48,51 +53,308 @@ def show(self, **kwargs):
...
+        plt.close()
+        C = cs.allsegs[0][0]
+        # convert this extracted contour to the shoreline mask
+        shoremap = np.zeros_like(data_trim)
+        flat_inds = list(map(lambda x: np.ravel_multi_index(x, shoremap.shape),
+                             np.fliplr(np.round(C).astype(int))))
+        shoremap.flat[flat_inds] = 1
+
+        # write shoreline map out to data.mask
+        self.mask = np.zeros(self.data.shape)
```
amoodie: I think you should set `mask` as a read-only property of the `BaseMask` class. Then it can be explicitly documented in the documentation (for someone looking to access the mask array specifically) and it would be sort of protected behind the `_`. 

```
@property
def mask(self):
    """ndarray : Binary mask values.

    Other relevant info....?
    """
    return self._mask
```

and then set it here with `self._mask[trim_idx:, :] = shoremap`

deltametrics/mask.py
```diff
@@ -48,51 +53,308 @@ def show(self, **kwargs):
...
+    """Identify a binary channel mask.
 
     A channel mask object, helps enforce valid masking of channels.
 
     Examples
     --------
-
     Initialize the channel mask
-        >>> cmsk = dm.mask.ChannelMask(arr)
+        >>> cmsk = dm.mask.ChannelMask(velocity, topo)
```
amoodie, line 277: this isn't completely clear. Add before this line:

```
>>> velocity = rcm8cube['velocity'][-1, :, :]
>>> topo = rcm8cube['eta'][-1, :, :]
```

deltametrics/mask.py
```diff
@@ -106,28 +368,199 @@ def a_channel_function(self):
...
+    shoreline computation to add the wet pixels on the topset back to the mask.
 
-    A land-water mask object, helps enforce valid masking of land water interface.
+    If a land mask has already been computed, then it can be used to define the
+    wet mask. Otherwise the wet mask can be computed from scratch.
 
     Examples
     --------
+    Initialize the wet mask
+        >>> wmsk = dm.mask.WetMask(arr)
```
amoodie, line 415: same as above, what exactly is `arr`? make it explicit by pulling from the `rcm8cube`

deltametrics/mask.py
```diff
@@ -48,51 +53,308 @@ def show(self, **kwargs):
...
+        # assign **kwargs to self
+        for key, value in kwargs.items():
+            setattr(self, key, value)
+
+        # set default values for **kwargs if they do not exist
+        self.velocity_threshold = getattr(self, 'velocity_threshold', 0.3)
+        self.topo_threshold = getattr(self, 'topo_threshold', -0.5)
+        self.angle_threshold = getattr(self, 'angle_threshold', 75)
+        self.numviews = getattr(self, 'numviews', 3)
+        self.is_mask = getattr(self, 'is_mask', False)
```
amoodie: was there a reason you chose this approach rather than `pop`ping, e.g., `self.velocity_threshold = kwargs.pop('velocity_threshold', 0.3)`? Moreover, I think `kwargs` is best used when you will pass the high-level function input `kwargs` down several chains of lower-level functions, but you don't want to have to specify all possible options along the way. It's also useful for interacting with external libraries like `matplotlib` so we don't have to explicitly handle all available options there.

In this case here, I think these are mostly actually parameters relevant for the OAM methods. So, do you really need to set them in each `xMask` or can you just look for them within the `OAM` method? This could help reduce some of the repetitive nature here. 

Similarly, do you need to use `self.is_mask` or can `is_mask=False` just be an optional parameter in the `__init__` function signature and then where you need it below on Line 320 just use it as `if is_mask:`?

deltametrics/mask.py
```diff
@@ -48,51 +53,308 @@ def show(self, **kwargs):
...
+
+        if self.is_mask is False:
```
amoodie: Why not just `if not self.is_mask:`?

deltametrics/mask.py
```diff
@@ -48,51 +53,308 @@ def show(self, **kwargs):
...
+            self.compute_channelmask()
+        elif self.is_mask is True:
+            self.mask = self.data
+        else:
+            raise ValueError('is_mask must be a `bool`,'
```
amoodie: did this get hit in tests? I think this will fail if self.is_mask` is not a string already.  

```
>>> a = True
>>> 'the test' + a
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: must be str, not bool
```

Could try fixing it with `str(self.is_mask)` I think.

deltametrics/mask.py
```diff
@@ -139,3 +572,320 @@ def a_land_function(self):
         """This is a wrapper to the below function.
         """
         return a_land_function(self.data)
+
+
+class ShoreMask(BaseMask, OAM):
```
amoodie: Would this be better named `ShorelineMask`? I don't feel strongly about this, just a thought

amoodie (changes requested):
Jay this is really great. I have made a few comments in the code that should be addressed. Several of these apply to each of the `xMask`s, but I only commented it on the first one.

Outstanding issues that should be addressed before merging:
* using the rcm8 cube as an example, the mask creation fails if there is no land above sea level (e.g., `dm.mask.ShoreMask(rcm8cube['eta'][0,:,:])`. This should probably return an empty (all zeros) mask? Add a test for this once you fix it please.
* remove the `time` property and `a_method` method from the `BaseMask`. These were just placeholders I put in when originally pushing the code.
* remove `property_for_just_channels` and `a_channel_function` from the `ChannelMask`. These were just placeholders.

Other issues that can be addressed later:
* as we have discussed, I think that there may be a simpler way to setup the whole masking ops. A lot of repetitive code and try-excepts to determine types of inputs and then create the necessary lower-level masks (e.g. `WetMask` and `LandMask`) before being able to create an `EdgeMask` could be eliminated by forcing specific lower-level mask types to be the inputs. For example, make the only arguments acceptable to `EdgeMask.__init__` one `WetMask` and one `LandMask` (or a single `WetMask`?). Then, the only acceptable input to `WetMask.__init__` is a `LandMask`, etc.  I think there would be some magic to do with the static methods then too, where each `xMask` has a `from_eta` static method, to create that `xMask` that would be a chain of the relevant lower-level masks.

deltametrics/mask.py
elbeejay: Good point, the line was covered in the tests, but the invalid value for `is_mask` was a string 'invalid' so it worked. I hadn't thought about using the `if not` convention but that is a change I can make. And yes I think the `str(self.is_mask)` will help cover cases with numerics or other types.

amoodie: cool. Yeah the safest bet is probably just `type(self.is_mask)`, because I think `str()` will also fail if the object does not have `__repr__` implemented.

deltametrics/mask.py
elbeejay: Swapped this for an attribute error

deltametrics/mask.py
elbeejay: Updated this for all of the mask code examples.

deltametrics/mask.py
elbeejay: Took care of this and all other mask code examples.

deltametrics/mask.py
elbeejay: Yep this worked to also help resolve the issue of creating a mask with an entirely submerged land mass.

elbeejay (commented):
> - using the rcm8 cube as an example, the mask creation fails if there is no land above sea level (e.g., dm.mask.ShoreMask(rcm8cube['eta'][0,:,:]). This should probably return an empty (all zeros) mask? Add a test for this once you fix it please.

The suggestion to amend the `__init__` to initialize the `_mask` object as zeros resolved this issue. A test has been added for this as well.

> - remove the time property and a_method method from the BaseMask. These were just placeholders I put in when originally pushing the code.

Cool, dropped these. 

> - remove property_for_just_channels and a_channel_function from the ChannelMask. These were just placeholders.

Dropped these as well.

I think most of your comments are addressed. The only thing that I am open to changing, but need to read about a bit, is the `kwargs` handling. I haven't used `pop` before so will look into that and where/when the `kwargs` is going to be useful.

deltametrics/mask.py
```diff
@@ -48,94 +46,823 @@ def show(self, **kwargs):
...
+        # set default values for **kwargs if they do not exist
+        self.velocity_threshold = getattr(self, 'velocity_threshold', 0.3)
+        self.topo_threshold = getattr(self, 'topo_threshold', -0.5)
+        self.angle_threshold = getattr(self, 'angle_threshold', 75)
+        self.numviews = getattr(self, 'numviews', 3)
+        self.is_mask = getattr(self, 'is_mask', False)
+
+        if not self.is_mask:
+            self.compute_channelmask()
+        elif self.is_mask is True:
```
amoodie: I shoudl have said explicitly: these should also be changed, as above, to `elif self.is_mask:`

elbeejay: Looking into this further and I think that switching to `elif self.is_mask:` means that any value for `is_mask` other than False is caught in this `elif` statement. I am also thinking that I might switch back to `False` instead of `if not` as I think it is more obvious what the loop is checking when `False` and `True` are explicitly stated (unless you have a strong objection to it).

amoodie: yeah okay, you're probably right: https://docs.quantifiedcode.com/python-anti-patterns/readability/comparison_to_true.html
Seems like `if cond is True:` is fine if you want to be sure it's boolean.

Sorry for the trouble.

amoodie (changes requested):
Hi @elbeejay 


## Errors
I'm getting errors on the following call still: `a = dm.mask.ChannelMask(rcm8cube['velocity'][0,:,:], rcm8cube['eta'][0,:,:])` and `tm = dm.mask.EdgeMask(rcm8cube['eta'][0,:,:])`. I guess this should probably return an empty mask, similar to the change you made for `ShorelineMask`.

The error is:
```
--> 734             self.landmask = (self.shore_image < self.angle_threshold) * 1
    735             self.wetmask = self.oceanmap * self.landmask

AttributeError: 'EdgeMask' object has no attribute 'shore_image'
```
so I guess it's probably something to do with a value not being set by the early-return from the ShorelineMask call. 

## Tests

I think a cleaner way to do the `test_plotter` is the following code:
```
def test_plotter(tmp_path):
    """Test the show() function."""
    import matplotlib.pyplot as plt
    shore_mask = mask.ShorelineMask(rcm8cube['eta'][-1, :, :])
    shore_mask.show()
    plt.savefig(tmp_path / 'mask_fig.png')
    plt.close()
    # assert that figure was made
    assert os.path.isfile(tmp_path / 'mask_fig.png')
```
This puts the file away in a temp folder, where it won't be accidentally committed. Your test would fail to remove the file if the test failed. There's also a pytest module for testing the actual look of matplotlib plots, which we may want to check out eventually (I used it on SedEdu stuff), but not necessary for now.

## Else

Looks awesome, great code, great tests, great docs :+1: :tada:

elbeejay:
@amoodie, thanks for the feedback!

---

> I'm getting errors on the following call still: a = dm.mask.ChannelMask(rcm8cube['velocity'][0,:,:], rcm8cube['eta'][0,:,:]) and tm = dm.mask.EdgeMask(rcm8cube['eta'][0,:,:]). I guess this should probably return an empty mask, similar to the change you made for ShorelineMask.

That ended up being pretty helpful. I realized I didn't do any type-checking or sanitation in the `BaseMask` class (or anywhere really). So I added checks to help enforce the `ndarray` input stipulation, or `Try-Except` to convert to an `ndarray` in the case of an object which has wrapped the `ndarray` class (such as our Cubes and presumably the Plans). Then added some tests to make sure the type sanitation and handling was happening correctly. 

---

> I think a cleaner way to do the test_plotter is the following code:

Had never heard of the `tmp_path` option for pytest so this was cool.

amoodie (approved):
looks great!

PR #8, opened 2020-07-02
Switching to GitHub Actions

elbeejay:
This workflow seems to function as desired and makes the switch from Travis CI to GitHub Actions + CodeCov for our CI. Tested it [here](https://github.com/elbeejay/deltametrics-actions) (will delete repo-copy after we merge this). 

**Note**: Before deploying a version to be built on PyPI we need to make the deploy key ("secret") for this project... otherwise I assume the deploy workflow will fail (need to call them PYPI_PASSWORD and PYPI_USERNAME).

### After Merging
- The badging images broke the build because the badges did not exist yet to reference. So after this is merged the badges can be added to the Readme.


amoodie (approved):
LGTM. Thanks for doing it. Aside, what is the "greetings" one? Those messages show up when someone opens an issue? I guess I'll see after merging this.

I'll merge this one now. Add the badges please, and PR and go ahead and merge it yourself.

elbeejay:
> Aside, what is the "greetings" one? Those messages show up when someone opens an issue?

The `greetings.yml` creates an automated message when someone creates their *first* issue or pull request. We could probably change that if we wanted it to be every issue/PR, but right now it is just for a user's first one. In the future we can also create default templates for issues and PRs if needed.

amoodie:
cool. Yeah I knew about templates, but not about the greeting thing. 

> In the future we can also create default templates for issues and PRs if needed.

Yeah, I'm not sure we're at that point yet :laughing:

PR #11, opened 2020-08-14
Stratigraphy overhaul/introduction

### LINKED ISSUE #6: Implement boxy stratigraphy
Implement the boxy stratigraphy routine in DeltaMetrics from pyDeltaRCM.

I think this should probably be the default stratigraphy we calculate and work with, with the option to compute mesh stratigraphy.

amoodie:
# Summary
This PR is a major overhaul of stratigraphy operations in the model. Herein, I introduce the `StratigraphyCube`, which is computed as 'boxy' stratigraphy, where each voxel in real z-x-y space has a data value. Notably, the computation for stratigraphy is fast, and only needs to occur once. After this initial computation, data are drawn Sections are readily displayed from both Cube types in multiple styles.

I've also made a ton of additions to the documentation and tests. Added a `NoStratigraphyError` which gets raised periodically. 

closes #6 

The todo list below is "complete". The items that are *not* checked, have issues open and will be addressed in the future.

### Preview image:
image

# Todo:
- [x] method to export a frozen z-x-y array of some variable from the `StratigraphyCube`
- [x] documentation `section lexicon` updates
- [ ] implementation of `BoxyStratigraphyAttributes` to match the `MeshStratigraphyAttributes`. Likely require significant refactoring of the `get_display_xxxx()` methods (see below).
- [x] move the `get_display_xxxx()` methods of the `DataSectionVariable` and `StratigraphySectionVariable` out to functions in `plot.py`. I think they probably make more sense to place the complicated if-then logic there, rather than part of the variables themselves, since the logic is all related to plotting.
- [ ] implement the `return_cube` options for functions in `strat.py`.
- [ ] related to above, need to implement the `__init__` options for `StratigraphyCube`s.
- [ ] implement options for `z`, `dz`, and `nz` at every entry point to the api, need to ensure the behavior is consistent, and a single default is set everywhere. Preferably a default would be set in only one location. Could make default `None` everywhere, and then if all `None`s into the `_determine_strat_coordinates()` function, set the default `dz` there?
- [x] add example of computing preserved vs total attributes
- [ ] fix off-by-one edge errors for the mesh display style 
- [x] fix `psvd` determined in the processing of `elev` by `strat._compute_elevation_to_preservation()`; misses partially preserved times.
- [x] fix mistake in coloring preserved timesteps, e.g., `np.array([0, 1, 3, 2, 4, 5])`; this might mean creating custom `fill_between()`.
- [x] robust testing suite


deltametrics/strat.py
```diff
@@ -32,3 +35,527 @@ def compute_compensation(line1, line2):
...
+    nz : :obj:`int`, optional
+        Number of intervals in `z`. Z array is created as 
+        ``np.linspace(np.min(elev), np.max(elev), num=nz, endpoint=True)``.
+    """
+    if (dz is None) and (z is None) and (nz is None):
+        raise ValueError('You must specify "z", "dz", or "nz.')
+
+    _valerr = ValueError('"dz" or "nz" cannot be zero or negative.')
+    if not (z is None):
+        if np.isscalar(z):
```
elbeejay, line 545: Could be more general with something like: `if type(z) is not numpy.ndarray`

deltametrics/strat.py
```diff
@@ -32,3 +35,527 @@ def compute_compensation(line1, line2):
...
+            raise ValueError('"z" must be a numpy array.')
+        return z
+    elif not (dz is None):
+        if dz <= 0:
+            raise _valerr
+        max_dos = np.max(elev) + dz  # max depth of section, meters
+        min_dos = np.min(elev)       # min dos, meters
+        return np.arange(min_dos, max_dos, step=dz)
+    elif not (nz is None):
+        if nz <= 0:
```
elbeejay, line 555: For this and `dz <= 0` - could be generalized to also check that the input is a number (rather than a string or something else).

deltametrics/strat.py
```diff
@@ -32,3 +35,527 @@ def compute_compensation(line1, line2):
...
+    """Compute the preserved timesteps.
+
+    The output from :obj:`_compute_elevation_to_preservation` records whether
+    an instance of time, defined exactly at the data interval, is recorded in
+    the stratigraphy (here, "recorded" does not include stasis). This differs
+    from determining which *time-intervals* are preserved in the stratigraphy,
+    because the deposits reflect the conditions *between* the save intervals.
+
+    While this computation is simply an offset-by-one indexing (``psvd[1:,
+    ...]``), the function is implemented explicityly and utilized internally
```
elbeejay: typo in 'explicitly'

docs/source/pyplots/guides/userguide_bed_timeseries.py
```diff
@@ -0,0 +1,13 @@
+rcm8cube = dm.sample_data.cube.rcm8()
+
+nt = 5
```
elbeejay, line 3: In user guide this is `nt=6` so will want to change either this line or the user guide `.rst` to be consistent.

elbeejay (approved):
This is definitely a big step for the `DeltaMetrics` framework. I'll admit I didn't have it in me to do a super in-depth review of the additions proposed, but I like the overall gist of things.

I anticipate any functionality that ends up being clunky or less than ideal will be reworked as we add additional components or change things. I think the base structure and setup of everything looks good. 

All in all, I am :astonished:

deltametrics/strat.py
amoodie: added note to #18

deltametrics/strat.py
amoodie: I think this would be less general actually... `None` is the default for all three options, so we check explicitly if it is not `None` here. Regardless, you are probably right that there is a better way to handle these three mutually-exclusive options with precedence case.

I will revisit in #18

amoodie:
great, thanks for the review @elbeejay. I will merge once the tests pass.

PR #23, opened 2020-09-06
Xarray support

elbeejay:
## Updated (9/28/2020)

This PR makes `CubeVariable` an `xarray`-accessor rather than a subclass of `numpy.ndarray`. So changes are all over the place, but are relatively minor and largely preserve the prior API.

### The changes

1. In `io.py` instead of reading the NetCDF4 as a dictionary, we load it as an `xarray.DataSet`. This retains the benefits of the original scheme ("lazy" loading without reading the file into memory). In this way we can keep the `.read()` function to give the user the option of fully reading the dataset into memory whenever they wish (presumably prior to running a serious computation). What is a bit clumsy right now is the definition of the "coords" which are the dimensions of the data arrays. This might be better addressed in `pyDeltaRCM` (related to https://github.com/DeltaRCM/pyDeltaRCM/issues/99) as there we could just create a better netCDF file to read (and for other datasets encourage people to preprocess or prep their datasets per the expected standard).

2. Turns out NetCDF4 and HDF5 are very similar (see [this](https://www.unidata.ucar.edu/software/netcdf/docs/interoperability_hdf5.html) and [this](https://www.unidata.ucar.edu/software/netcdf/docs/faq.html)). The built-in `xarray` loading supports most NetCDF4 and HDF5 files so now we use that and have therefore added support for HDF5 files. A small HDF5 example file has been added to allow us to unit test this functionality (as well as generalize DeltaMetrics by incorporating example Landsat data).

3. The developers of `xarray` have been pretty consistent (back from when it was `xray` even), that they do not support or encourage subclassing (their issues [728](https://github.com/pydata/xarray/issues/728), [3959](https://github.com/pydata/xarray/issues/3959), and [3980](https://github.com/pydata/xarray/issues/3980) to name a few). Instead they suggest using an "accessor" to extend the functionality of `xarray` objects. So now `CubeVariable` is an accessor to `xarray` allowing us to add custom functions to `CubeVariable`, but retain an `xarray` object as `CubeVariable.data`.

4. Rudimentary 3D plotting support added for `show_cube()` via [PyVista](https://github.com/pyvista/pyvista). Returns an error if `pyvista` is not installed.

### Conclusions

This passes the unit tests as well as the documentation tests which suggests that this change does not break everything. But I do suspect that some of the stratigraphy functions are a bit less stable than they were before due to changes in the way some of the dimensions (think `self.x` or `self._y` etc) are handled. Started tracking things down, but realized that we might want to convert components of `plan.py` and `section.py` to be `xarray` accessors too (such as `SectionVariable`). So stopped messing with things and have settled for the PR as it is now. Implementing this might just be an incremental step towards true `xarray` compatibility, details regarding coordinates and dimensions are incomplete as mentioned above.


deltametrics/cube.py
```diff
@@ -506,13 +523,27 @@ def __getitem__(self, var):
         CubeVariable : `~deltametrics.cube.CubeVariable`
             The instantiated CubeVariable.
         """
-        if var == 'time':
-            # a special attribute we add, which matches eta.shape
-            _t = np.expand_dims(self.dataio['time'], axis=(1, 2))
-            return CubeVariable(np.tile(_t, (1, *self.shape[1:])),
-                                variable='time')
+        if var == 'time' or var == 'x' or var == 'y':
```
amoodie: `if var in self.known_coords`? Not certain whether that is faster or not tho...

deltametrics/strat.py
```diff
@@ -378,19 +381,40 @@ def _compute_elevation_to_preservation(elev):
...
+    else:
+        strata[-1, ...] = elev.data.values[-1, ...]
+        for j in np.arange(nt - 2, -1, -1):
+            strata[j, ...] = np.minimum(elev.data.values[j, ...],
+                                        strata[j + 1, ...])
+            psvd[j + 1, ...] = np.less(strata[j, ...],
+                                       strata[j + 1, ...])
+        if nt > 1:  # allows a single-time elevation-series to return
+            psvd[0, ...] = np.less(strata[0, ...],
+                                   strata[1, ...])
```
amoodie: this is messy and will be very annoying to maintain. Can we just coerce whatever the input `elev` is into a numpy array and then use that? Even if you just said something like (pseudocode)

```python
if isinstance(elev, np.ndarray):
    _elev = elev
elif isinstance(elev, xr.core.dataarray.DataArray):
    _elev = elev.values
```
before the computation, and then used `_elev` inside the computation, that would be much much better.

deltametrics/section.py
```diff
@@ -301,16 +301,24 @@ def __getitem__(self, var):
         """
         if type(self.cube) is cube.DataCube:
             if self.cube._knows_stratigraphy:
-                return DataSectionVariable(_data=self.cube[var][:, self._y, self._x],
+                return DataSectionVariable(_data=self.cube[var].data.values[:,
+                                                                     self._y,
+                                                                     self._x],
```
amoodie, line 306: I guess this is some of the stuff that is "clumsy" right now? I agree there may be a better way to do this, but I don't know it for sure.

deltametrics/cube.py
```diff
@@ -17,13 +19,15 @@
 from . import plot
 
 
-class CubeVariable(np.ndarray):
+@xr.register_dataarray_accessor("cubevar")
+class CubeVariable():
```
amoodie, line 23: I think some of the clunkiness can be taken out by adding a `__getitem__` to this class. Adding the following makes it so that you don't have to do `rcm8cube['eta'].data[10,:,:]`, but can just keep the straightforward API we have in place (i.e., `rcm8cube['eta'][10,:,:]`) and the object you get back is still a proper `xarray` `DataArray`. I think this will be something we definitely want to straighten out before merging. This should simplify a lot of additional `.data` calls.

```python
    def __getitem__(self, slc):
        return self.data[slc]
```

amoodie (changes requested):
Hi @elbeejay. This was a huge effort and you have pushed the project way forward with it, thank you.

My only major suggestion is adding a `__getitem__` to the `CubeVariable` that points to the `xarray` data. This allows the API for accessing data to stay the same, and would remove a lot of the needed `.data` calls (for example in the `mask` codes).

deltametrics/cube.py
elbeejay: Good idea, we store the coordinates in the BaseCube as `self._coords` so went with that.

deltametrics/section.py
elbeejay: Yes it feels like there should be a better way... maybe we'll happen upon it as we move towards more `xarray` objects and start actually using the coordinates etc.

deltametrics/strat.py
elbeejay: Yes I agree. That chunk of code is a bit embarrassing to look at

elbeejay:
I think between your commits @amoodie and some of the smaller commits I made today we've made this into a better PR. In the future I'm sure there are many more `xarray` related methods/PRs that'll come up, but I think this is enough to get us started on the path of "easy" dimensioning for our plots/computations...

amoodie:
Cool, thanks for cleaning up. I agree, there will be more to do, but this is a huge first step. Thanks!

amoodie (approved):

PR #26, opened 2020-09-15
Section traces

amoodie:
Develop new section types. Principally add effective `PathSection`, and the `CircularSection` and `RadialSection` types. There is still more to do here, but I have opened issues for the main points.

# Todo
- [x] finish implementation for RadialSection angles
- [x] finish tests for RadialSection
- [x] implement the `DipSection` or open an issue
- [ ] can the optional argument be streamlined better (simpler)? Open an issue
- [x] section arguments need to work for actual dimensions of the arrays (`xarray`) not just indices. Open an issue to address this after Jay's #23

### DIFF
--- a/deltametrics/section.py
+++ b/deltametrics/section.py
@@ -178,14 +178,15 @@ class BaseSection(abc.ABC):
 
     """
 
-    def __init__(self, section_type, *args):
+    def __init__(self, section_type, *args, name=None):
         """
         Identify coordinates defining the section.
 
         Parameters
         ----------
         CubeInstance : :obj:`~deltametrics.cube.Cube` subclass instance, optional
-            Connect to this cube. No connection is made if cube is not provided.
+            Connect to this cube. No connection is made if cube is not
+            provided.
 
         Notes
         -----
@@ -202,12 +203,14 @@ class BaseSection(abc.ABC):
         self._trace = None
         self._shape = None
         self._variables = None
+        self._name = name
         self.cube = None
 
         self.section_type = section_type
 
         if len(args) > 1:
-            raise ValueError('Expected single argument to %s instantiation.'
+            raise ValueError('Expected single positional argument to \
+                             %s instantiation.'
                              % type(self))
 
         if len(args) > 0:
@@ -225,7 +228,7 @@ class BaseSection(abc.ABC):
                                 _gottype=type(CubeInstance)))
         self.cube = CubeInstance
         self._variables = self.cube.variables
-        self._name = name or self.section_type
+        self._name = self._name or name or self.section_type
         self._compute_section_coords()
         self._compute_section_attrs()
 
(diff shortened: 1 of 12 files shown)


deltametrics/section.py
```diff
@@ -223,9 +228,14 @@ def connect(self, CubeInstance):
                                 _gottype=type(CubeInstance)))
         self.cube = CubeInstance
         self._variables = self.cube.variables
+        self._name = self._name or name or self.section_type
```
elbeejay: Is this the intended hierarchy for Section name assignment? This would mean that a named section that is connected to a new `CubeInstance` maintains its old name even if a new one is supplied right? I can see why that'd be good, but I can also imagine it could be confusing if you supply a `name` argument and it isn't used (although putting `name` first would make it `None` quite often I suppose)...

deltametrics/section.py
```diff
@@ -518,10 +703,268 @@ def __init__(self, x=-1):
...
+        self._input_radius = radius
+        self._input_origin = origin
+        super().__init__('circular', *args, **kwargs)
+
+    def _compute_section_coords(self):
+        if (self._input_radius is None):
+            if not (self.cube is None):
+                self.radius = int(np.min(self.cube.shape[1:]) / 2)
+            else:
+                self.radius = 1
```
elbeejay: Just out of curiosity, did you have any scenarios in mind where this condition might occur?

deltametrics/section.py
```diff
@@ -518,10 +703,268 @@ def __init__(self, x=-1):
...
+        if (self._input_azimuth is None):
+            self.azimuth = 90
+        else:
+            self.azimuth = self._input_azimuth
+
+        # determine the origin of the line
+        if (self._input_origin is None):
+            if not (self.cube is None):
+                land_width = utils.guess_land_width_from_land(
+                    self.cube['eta'][-1, :, 0])
```
elbeejay: I think this should be consistent with the method used in `CircularSection`. Right now this implementation is putting the default origin on the opposite side of the domain when applied to a `StratigraphyCube` made from the example data. (Implementation for `CircularSection` is working as expected)

Code snippet if you want to see what I mean:
```
sc8cube = dm.cube.StratigraphyCube.from_DataCube(rcm8cube)
sc8cube.register_section('radial', dm.section.RadialSection(azimuth=45))
# show the location and the "velocity" variable
fig, ax = plt.subplots(2, 1, figsize=(8, 4))
rcm8cube.show_plan('eta', t=-1, ax=ax[0], ticks=True)
sc8cube.sections['radial'].show_trace('r--', ax=ax[0])
sc8cube.sections['radial'].show('velocity', ax=ax[1])
plt.show()
```

deltametrics/utils.py
```diff
@@ -218,3 +201,280 @@ def func_linear(x, a, b): return a * x + b
     perror = np.sqrt(np.diag(pcov))
 
     return yfit, pcov, perror
+
+
+def guess_land_width_from_land(land_col_0):
```
elbeejay, line 206: I know this is a 'temporary' method until we add the land width to the netCDF file, but could this be called during cube initialization and held in the cube as `self._trim_idx` or something similar? Might make the swap down the line easier to implement as the function would be called once during initialization instead of spread across multiple files/methods?

elbeejay (commented):
Nice stuff, the passing of the *args/**kwargs for the plotting is handy to have. 

Ran into a minor issue with the origin of the radial section, but hopefully that is minor. Similarly I had a small suggestion about the `L0` identification method and where/how it could be used.

deltametrics/section.py
amoodie: thanks, good catch. Ultimately, this will be reworked after the netcdf file is updated to include `L0`

deltametrics/utils.py
amoodie: Probably a good idea, and related to my bug on the `RadialSection` too. I'm going to defer to after updating the netcdf file though, to see what the best way to pull this info into the `Cube` is. I think probably creating some attributes of the cube based on whether those attributes exist in the netcdf file will make the most sense. For example, if `L0` is in the netcdf, then make `self.L0` in the `Cube` and then when creating a `Section`, you can check whether `Cube.L0` exists, and use it if it does. 

I'm going to open an issue for this.

deltametrics/section.py
amoodie: When I was writing it, I thought it might happen, but then when I was testing, I realized that `_compute_section_coords` is never called until a cube is connected with `.connect()`. So no, I don't think this branch is ever going to be hit. I'm going to remove these branches from all the `Sections`.

deltametrics/section.py
amoodie: Yeah, that seems like it would make sense though, like the `_name` is meant to mean that it is protected attribute. I could make a property `@name` with this same logic. At least then I could issue a warning in your mentioned scenario...

Let me look at it.

amoodie: Cool, I gave `_name` a setter, so that a warning is issued if you try to give a name to an already named section.

PR #27, opened 2020-09-29
let masks be 3d if they want to be

### LINKED ISSUE #20: Masks currently locked to sea level == 0
Are masks currently locked to sea level == 0? I think they might be. This should be relaxed, at least with an option to specify during mask creation. 

related to #10. Should consider how we will include boundary conditions in an overhaul of the netcdf file output from pyDeltaRCM too.

elbeejay:
The masking functions were previously written in such a way that they *required* a 2-D xy slice of the data. This seemed like a poor requirement to have, as most people will likely want to mask their data over the duration of the experiment/simulation.

### This PR does the following:

1. Masks are more flexible, you can still pass in 2-D data if you want (e.g. `landmask = mask.LandMask(cube['eta'][50, :, :])`) but now you can also pass in a 3-D variable and do something like `landmask = mask.LandMask(cube['eta'])`.

2. The in-built plotting function of the masks has an optional argument for the time slice you want to display, so `landmask.show(5)` displays the mask corresponding to `landmask[5, :, :]`. By default the *final* mask is shown (so the index value -1 is used)

3. Some unit tests were added to check that 3-D data works for all the masks (note: backwards compatibility was not broken so old tests were untouched) 

### This PR doesn't:

- Establish a `register_mask` method for the general data cube

- Speed up the mask functions (there is a lot of room for improvement)

- Do anything related to #20

Despite these shortfalls, this is still a step forward I think, and allows us to think about writing functions for surface metrics that rely on binary mask information through time (e.g. channel mobility).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -48,10 +48,10 @@ jobs:
       uses: actions/checkout@v2
       with:
         persist-credentials: false
-    - name: Set up Python 3.x
+    - name: Set up Python 3.8
       uses: actions/setup-python@v2
       with:
-        python-version: '3.x'
+        python-version: '3.8'
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
--- a/deltametrics/mask.py
+++ b/deltametrics/mask.py
@@ -25,6 +25,16 @@ class BaseMask(object):
         else:
             self.data = data
 
+        # set data to be 3D even if it is not (assuming t-x-y)
+        if len(np.shape(self.data)) == 3:
+            pass
+        elif len(np.shape(self.data)) == 2:
+            self.data = np.reshape(self.data, [1,
+                                               np.shape(self.data)[0],
+                                               np.shape(self.data)[1]])
+        else:
+            raise ValueError('Input data shape was not 2-D nor 3-D')
+
         self._mask = np.zeros(self.data.shape)
 
     @property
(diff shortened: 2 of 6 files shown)


amoodie (approved):
Hi @elbeejay 

This looks fine, I think. I understand the reasoning here is to have something consistent with the cube variables, which I like. My only question is whether it is necessary to expand a single-time mask to three dimensional? Would this be confusing for use with other functions? I guess it makes it simpler for our internals (like `mask.show`).

elbeejay:
> My only question is whether it is necessary to expand a single-time mask to three dimensional? Would this be confusing for use with other functions? I guess it makes it simpler for our internals (like `mask.show`).

Yeah I wasn't sure the best way to handle that. I made them all 3-D for convenience when it came to internal functions, testing, etc. as then the shape of the Mask is always 3-D. Happy to change it and have a 2-D case / conditionals for the single-time masks.

amoodie:
I think it's probably okay. I just worry about someone trying to operate with the mask they get back and being confused why it has a weird shape. Maybe you can just add some clarification to this in the documentation for now, if we need to change it back later, that's fine. After all, for `lm = mask.LandMask(cube['eta'][50, :, :])` your code does actually return an obejct with the same dimensions... So, you could also say in the docs something like:

```
.. important::

    When `Mask` objects are instantiated with single time-slice x-y planform
    data (i.e., `L x W` data), the returned mask is expanded along the first
    dimension to return an object with dimensions `1 x L x W`. This ensures
    internal consistency in the project, but can lead to confusion if writing
    your own functions. 
```

elbeejay:
> I think it's probably okay. I just worry about someone trying to operate with the mask they get back and being confused why it has a weird shape. Maybe you can just add some clarification to this in the documentation for now, if we need to change it back later, that's fine. 

Sounds good, I've added a note to the `mask.py` API documentation for now.

amoodie:
cool, I think this will pass all the tests, but one failed on coverage. I'll merge on green.

PR #28, opened 2020-10-04
Channel mobility metrics

elbeejay:
This would go after #27.

## Summary
This PR provides at least an initial take at adding channel mobility metrics to the `deltametrics` framework. Nothing fancy was done, so we can have a discussion about tying these functions in more tightly to `deltametrics` objects (right now they are much more 'standalone' than other modules we have). But the 4 metrics added are the following:

1. Dry fraction decay per Cazanacli et al 2002

2. Channel Planform Overlap per Wickert et al 2013

3. Reworking Fraction per Wickert et al 2013

4. Channel Abandonment per Liang et al 2016

A demo of the current functionality is available [here](https://colab.research.google.com/drive/1-lkswSZSGRkhsArLm245WMSiKvj03YyU?usp=sharing). This PR does not include any substantial documentation beyond linking up the API reference, but in the future the plan would be to create documentation resembling that colab notebook.

## Details
More specifically, this PR adds the following:
- Allows for flexible type inputs to the mobility metrics
- Functions to compute the above mobility metrics
- Function for curve fitting linear, harmonic, and exponential functions
- Unit tests the methods against the simple 'demo' test in addition to standard error/exception checking
- Updates api reference to document these functions

This PR does **not** do the following:
- Add an example of using these metrics to the documentation
- Provide these functions as any kind of deltametrics class or object
- Create a new 'base' class of any kind
- Implement cool function names (I don't know that `calc_`"metric name here" is the nicest way to name a function)
- Add methods to create more general (geometric) base masks, this could be something that is added to the masking functions later (e.g. a method to create binary mask of a radial region).
- Manage the time dimension (these methods use indices from the arrays for setting `basevalues` and `time_windows`), in the future we could build-in the time 'units' via `xarray`

So like I said, I'm very flexible on changing the structure of these functions to better mesh with the deltametrics API, it was just not obvious to me what the best way of doing that would be. Part of my hesitation about writing a complex API / additional documentation is just due to the uncertainty around the merge order (will this go before or after #23 which breaks some bits of the API, either way I think the conflicts can be resolved without too much fuss). In conclusion, despite the shortcomings of this PR, I think it is still a step forward.

### DIFF
--- a/deltametrics/mobility.py
+++ b/deltametrics/mobility.py
@@ -346,8 +346,10 @@ def calc_chan_abandonment(chmap, basevalues, time_window):
     Returns
     -------
     PwetA : ndarray
-        Vector of number of pixels that are abandonded for a given time lag,
-        averaged over the time window under study
+        A 2-D array of the abandoned fraction of the channel over the window of
+        time. It is of shape len(basevalues) x time_window so each row
+        represents the fraction of the channel that has been abandonded, and
+        the columns are associated with each time lag.
 
     """
     # sanitize the inputs first
@@ -355,7 +357,7 @@ def calc_chan_abandonment(chmap, basevalues, time_window):
                                                            basevalues,
                                                            time_window)
     # initialize values
-    PwetA = np.zeros((time_window, 1))
+    PwetA = np.zeros((len(basevalues), time_window))
     chA_base = np.zeros_like(chmap[0, :, :])
     chA_step = np.zeros_like(chmap[0, :, :])
 
@@ -366,20 +368,14 @@ def calc_chan_abandonment(chmap, basevalues, time_window):
         chA_base = chmap[Nbase, :, :]
         # get total number of channel pixels in that map
         baseA = np.sum(chA_base)
-
         # loop through the other maps to be compared against the base map
         for Nstep in range(1, time_window):
             # get the incremental map
             chA_step = chmap[Nbase+Nstep, :, :]
-            # get the number of pixels that are no longer channelized (aka
-            # abandonded)
+            # get the number of pixels that were abandonded
             stepA = len(np.where(chA_base.flatten() > chA_step.flatten())[0])
-            # store this number in the vector PwetA for each incremental map
-            PwetA[Nstep] = PwetA[Nstep] + (1-stepA/baseA)
-
-    # normalize PwetA by the number of incremental steps used to generate each
-    # value
-    PwetA = PwetA/time_window
+            # store this number in the PwetA array for each transient map
+            PwetA[i, Nstep] = stepA/baseA
 
     return PwetA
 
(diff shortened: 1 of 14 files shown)


deltametrics/mobility.py
```diff
@@ -0,0 +1,453 @@
...
+    # check that the combined basemap + timewindow does not exceed max t-index
+    Kmax = np.max(basevalues) + time_window
+    if Kmax > chmap.shape[0]:
+        raise ValueError('Largest basevalue + time_window exceeds max time.')
+
+    # return the sanitized variables
+    return chmap, landmap, basevalues, time_window
+
+
+def calc_chan_decay(chmap, landmap, basevalues, time_window):
```
amoodie: use full name `calculate_channel_decay`

deltametrics/mobility.py
```diff
@@ -0,0 +1,453 @@
...
+            chA_step = chmap[Nbase+Nstep, :, :]
+            # get the number of pixels that were abandonded
+            stepA = len(np.where(chA_base.flatten() > chA_step.flatten())[0])
+            # store this number in the PwetA array for each transient map
+            PwetA[i, Nstep] = stepA/baseA
+
+    return PwetA
+
+
+def mobility_curve_fit(mdata, fit='harmonic'):
```
amoodie: Is there any delta-specific or mobility-specific about this? If we want to include some simple curve-fitting utilities, I think they would fit better in `utils.py` and just be as generic as possible.

docs/source/reference/mobility/index.rst
```diff
@@ -0,0 +1,30 @@
...
+
+.. mobility functions
+.. ===================
+
+.. currentmodule:: deltametrics.mobility
+
+.. autosummary::
+	:toctree: ../../_autosummary
+
+  check_inputs
```
amoodie, line 25: spacing?

amoodie (approved):
Hi @elbeejay 

This is really great, thanks for contributing it! The notebook is a cool way to show it off, too.

I am wondering about the name of the module though (`mobility`). I'm not objecting, just trying to think about other metrics we will add. Probably a handful for shoreline mobility. Channel network metrics. Shoreline position/shape metrics. I think these all fall into something like `planform metrics` but we already have the `Planform` object housed in `plan.py`, and I'd like to keep those separate. I supposed `mobility` is probably broad enough it could encompass enough functions to make a complete module, but where do the shoreline_position metrics go then? Just thinking out loud here.

Please open issues for the items in your "does not do" checklist, where appropriate.

deltametrics/mobility.py
elbeejay: That's a good point, this is pretty generic curve-fitting so it doesn't need to be grouped with these functions.

docs/source/reference/mobility/index.rst
elbeejay: Thanks for catching this :+1:

deltametrics/mobility.py
elbeejay: They've been moved to `utils.py`

PR #36, opened 2020-10-29
fix time decoding in xarray

amoodie:
'time' in the netcdf file was being decoded to nano-seconds precision `HH:MM:SS` style time from `xarray.open_dataset()` method. We add an option to not decode times when opening the dataset here, which fixes the time to remain as just a numeric.

### DIFF
--- a/deltametrics/io.py
+++ b/deltametrics/io.py
@@ -166,7 +166,7 @@ class NetCDFIO(BaseIO):
             _tempdataset.close()
 
         try:
-            _dataset = xr.open_dataset(self.data_path)
+            _dataset = xr.open_dataset(self.data_path, decode_times=False)
 
             if 'time' and 'y' and 'x' in _dataset.variables:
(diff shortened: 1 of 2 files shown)


elbeejay (approved):
Works for me. Should still be flexible for other datasets, just is a formatting change right?

amoodie:
Yeah, I think so. It basically just prevents the `time` field from being converted to a `np.datetime64[ns]` data type. The problem was that we record in seconds, and it was being converted to nanoseconds. This just removes any conversion. Might be a more flexible way to do this from the netCDF writing angle though....maybe I should look at that first.

amoodie:
Alright, I found a better way to do this, in changing the output netcdf from pyDeltaRCM. Basically, naming the netcdf `variable` `unit` as `second` rather than `seconds` makes xarray not convert to a weird datetime thing and just keep it as a numeric. `second` is the preferred netcdf convention. Overall, this will allow netcdf files with actual datetimes to still be properly parsed.

Thanks for your question, closing this PR.

Issue #31, opened 2020-10-16
Generic (geometric) mask creation


elbeejay:
Add more generic masking functions, such as options for radial masks, rectangular masks, circular masks etc. (From #28)

elbeejay:
This exists, see #38

PR #38, opened 2020-11-05
Geometric masks

elbeejay:
Some basic geometric masking functions.

Right now the angular mask is not as flexible as I'd like so there's a note in the doc-string with what will eventually be improved. Will open up issues to create some documentation for these new masks and to fix the angular mask. But basically you can define and combine masks to get something like this:

image

Also adds a `channel presence()` function which returns an array of normalized channel presence (fraction of time channelized) given a 3-D array of channel masks. This is a pretty popular visualization of DeltaRCM results. 

Relates to #31.

### DIFF
--- a/deltametrics/mask.py
+++ b/deltametrics/mask.py
@@ -978,14 +978,34 @@ class GeometricMask(BaseMask):
                             'but was: ' + str(type(is_mask)))
 
         _, self.L, self.W = np.shape(self._mask)
-        self.xc = int(self.L/2)
+        self.xc = 0
         self.yc = int(self.W/2)
 
-    def angular(self, theta1, theta2):
+    @property
+    def xc(self):
+        """x-coordinate of origin point."""
+        return self._xc
+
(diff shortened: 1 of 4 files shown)


amoodie (approved):
This is cool, and code looks fine. What is the use case for these though? Like determining a region of interest, I'd guess?

Looks good to me :+1:

elbeejay:
> This is cool, and code looks fine. What is the use case for these though? Like determining a region of interest, I'd guess?

Use case would be exactly that, analyzing a specific region of interest. For example you could mask out the inlet channel and proximal region before calculating channel mobility metrics if you were interested in the mobility of the distal region. Or in the case of local subsidence you might want to mask out the subsiding region and compute metrics over there, and then compute metrics over the rest of the delta. 

The ability to set a circular mask of a specified radius at any arbitrary location is maybe a bit unncessary, but maybe it'll help answer a question one day... :man_shrugging:

Issue #10, opened 2020-07-07
Refactoring mask.py code


elbeejay:
Per some of the discussion in #7, there is a need to refactor some of the masking code.

Some suggestions from @amoodie in #7: 

> as we have discussed, I think that there may be a simpler way to setup the whole masking ops. A lot of repetitive code and try-excepts to determine types of inputs and then create the necessary lower-level masks (e.g. WetMask and LandMask) before being able to create an EdgeMask could be eliminated by forcing specific lower-level mask types to be the inputs. For example, make the only arguments acceptable to EdgeMask.__init__ one WetMask and one LandMask (or a single WetMask?). Then, the only acceptable input to WetMask.__init__ is a LandMask, etc. I think there would be some magic to do with the static methods then too, where each xMask has a from_eta static method, to create that xMask that would be a chain of the relevant lower-level masks.

There is also room to make the shoreline detection method more flexible. There is currently a built-in assumption of some *land width* in the input (like DeltaRCM) and automated trimming of the data prior to masking is reliant on this assumption (Lines [101](https://github.com/elbeejay/DeltaMetrics/blob/dc2ab7792a9842f10426dfe45a17d2481c9c6f3f/deltametrics/mask.py#L101) to [109](https://github.com/elbeejay/DeltaMetrics/blob/dc2ab7792a9842f10426dfe45a17d2481c9c6f3f/deltametrics/mask.py#L109)). It would be good to find an alternative solution so that these methods work for non-DeltaRCM data as well.

amoodie:
This is resolved since #44.

PR #44, opened 2021-03-29
Refactor masks, add basic Planform, add class OpeningAnglePlanform

amoodie:
# summary
This PR significantly refactors the masks in DeltaMetrics. **It is a breaking change, with respect to the dimensionality of masks!!**

# notable changes
* Ports the opening angle method out to a function in `plan` and a wrapper called the `OpeningAnglePlanform` which is a subclass of the `BasePlanform` type. These classes/ functions handle the OAM.
*Masks can now be created from one another and from an `OAP`, making it much faster to create multiple masks for one time slice.

# highlight figures

`OpeningAnglePlanform`
image

`Mask` hightlight
image

# Todo

- [x] add testing for new mask types
- [x] add testing for planform objects
- [x] fix connection and testing for mobility
- [x] fix the landmask to include the background land area. Without this, it creates some new bugs in picking the shoreline (it becomes a full circle rather than a line) and this leads to problems with shoreline length and roughness.
- [x] add _basic_ documentation for masks, planforms, and `OpeningAnglePlanform`. 
- [x] open an issue to develop better documentation for mask relationships and wrt `OpeningAnglePlanform`.
- [x] open issue to complete implementations for feeding masks directly (this might be best handled by relocating some of the static method infrastructure to the `__init__` of a mask, and then masking the static methods just wrap with type checking for inputs...?)
- [x] open issue to improve error messages throughout the masking and planforms. Many of these are open, unelaborated, errors.

# other notes
* adds a helpful error if you try to pass 3d arrays, i.e., to make you aware that this was a breaking change with respect to dimensionality of masks.
* deprecates the `is_mask` argument, in favor of static methods `from_array` for each mask type.

### DIFF
--- a/deltametrics/mobility.py
+++ b/deltametrics/mobility.py
@@ -46,6 +46,10 @@ def check_inputs(chmap, basevalues, time_window, landmap=None):
         chmap = chmap.mask
     elif isinstance(chmap, xr.core.dataarray.DataArray) is True:
         chmap = chmap.values
+    elif isinstance(chmap, list):
+        # assume it is a timeseries of masks set into a list
+        _arrs = [msk._mask.astype(np.int) for msk in chmap]
+        chmap = np.array(_arrs)
     else:
         raise TypeError('chmap data type not understood.')
 
@@ -61,6 +65,10 @@ def check_inputs(chmap, basevalues, time_window, landmap=None):
             landmap = landmap.mask
         elif isinstance(landmap, xr.core.dataarray.DataArray) is True:
             landmap = landmap.values
+        elif isinstance(landmap, list):
+            # assume it is a timeseries of masks set into a list
+            _arrs = [msk._mask.astype(np.int) for msk in landmap]
+            landmap = np.array(_arrs)
         else:
             raise TypeError('landmap data type not understood.')
         if ((landmap == 0) | (landmap == 1)).all():
--- a/deltametrics/mobility.py
+++ b/deltametrics/mobility.py
@@ -42,7 +42,10 @@ def check_inputs(chmap, basevalues, time_window, landmap=None):
     # check binary map types - pull out ndarray from mask or xarray if needed
     if isinstance(chmap, np.ndarray) is True:
         pass
-    elif isinstance(chmap, mask.BaseMask) is True:
+    elif issubclass(type(chmap), mask.BaseMask) is True:
+        raise TypeError(
+            'Cannot input a Mask directly to mobility metrics. '
+            'Use a list-of-masks instead.')
         chmap = chmap.mask
     elif isinstance(chmap, xr.core.dataarray.DataArray) is True:
         chmap = chmap.values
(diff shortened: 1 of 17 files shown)


amoodie:
@elbeejay maybe you can just scan this over, and give any high-level feedback? I don't expect you to verify everything.

The documentation and testing are both incomplete, tbh, but I need to move onto other things for the time being.

Let me know what you think!

amoodie:
@elbeejay I also think this closes #10, but wonder what you think.

deltametrics/mobility.py
```diff
@@ -46,6 +46,10 @@ def check_inputs(chmap, basevalues, time_window, landmap=None):
         chmap = chmap.mask
```
elbeejay, line 49: If masks are now 2-D arrays, then checking and allowing a `mask.BaseMask` to be used for mobility is not going to work anymore. I think either this `elif` could be removed, or it should throw an error.

deltametrics/plan.py
```diff
@@ -2,7 +2,383 @@
...
+    This object should wrap around many of the functions available from
+    :obj:`~deltametrics.mask` and :obj:`~deltametrics.mobility`.
+    """
+
+    def __init__(self, planform_type, *args, name=None):
+        """
+        Identify coordinates defining the planform.
+
+        Parameters
+        ----------
```
elbeejay, line 35: Should add to the doc-string to specify what `planform_type` and `name` are and their expected types.

deltametrics/plan.py
```diff
@@ -2,7 +2,383 @@
...
+        """
+        return self._shape
+
+
+class OpeningAnglePlanform(BasePlanform):
+    """Planform for handling the Shaw Opening Angle Method.
+
+    This `Planform` (called `OAP` for short) is a wrapper/handler for the
+    input and output from the :func:`shaw_opening_angle_method`. The `OAP` is a
+    conventient way to manage the extraction of a shoreline or a delta topset
```
elbeejay: typo in "convenient"

elbeejay (approved):
This looks good to me as a starting point for the `BaskMask` and `BasePlanform` layout and structure. I figure any/all of this code is subject to change as we move towards a more finished product.

So with that in mind feel free to skip over my handful of comments, maybe create issues if any seem worthy of one. Otherwise I don't think they are a big deal as those sections of code are likely to be revisited and edited in the future anyway.

deltametrics/mobility.py
amoodie: Ah yeah, I see that now, I've changed the statement in here so that any subclass of `BaseMask` as input will raise an error:
```
TypeError: Cannot input a Mask directly to mobility metrics. Use a list-of-masks instead.
```

amoodie: also added a test to check this.

deltametrics/plan.py
amoodie: fixed.

PR #50, opened 2021-04-05
Add support for 'sandfrac'

elbeejay:
Adds native support for the `sandfrac` variable without removing support for `strata_sand_frac`. A full refactor should take place in the future to update example datasets and remove built-in support for `strata_sand_frac`.


amoodie (approved):
Looks good, thanks for doing this, I was going to have to do it in a few days!

PR #55, opened 2021-04-28
Irregular-spaced shoreline position computation

amoodie:
A function to compute the mean and std dev to the shoreline. This method uses irregular sampling, which makes it relatively straightforward and computationally efficient, but it may not always match user intuition (for example in the presence of an estuary).

I have opened an issue to create a companion function that implements a regular sampling approach (#57).

Todo
 - [x] docs
 - [x] tests

### DIFF
--- a/deltametrics/plan.py
+++ b/deltametrics/plan.py
@@ -757,6 +757,20 @@ def compute_shoreline_distance(shore_mask, origin=[0, 0],
 
     Examples
     --------
+
+    .. doctest::
+
+        golf = dm.sample_data.golf()
+
+        sm = dm.mask.ShorelineMask(
+            golf['eta'][-1, :, :],
+            elevation_threshold=0,
+            elevation_offset=-0.5)
+
+        # compute mean and stddev distance
+        mean, stddev = dm.plan.compute_shoreline_distance(
+            sm, origin=[golf.meta['CTR'].data, golf.meta['L0'].data])
+
     """
     # check if mask or already array
     if isinstance(shore_mask, mask.ShorelineMask):
--- a/deltametrics/plan.py
+++ b/deltametrics/plan.py
@@ -748,12 +748,16 @@ def compute_shoreline_distance(shore_mask, origin=[0, 0],
 
     Returns
     -------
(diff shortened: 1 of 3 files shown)


deltametrics/plan.py
```diff
@@ -717,6 +717,85 @@ def compute_shoreline_length(shore_mask, origin=[0, 0], return_line=False):
...
+        ``Nx2`` array of x-y points is returned. Default is `False`.
+
+    Returns
+    -------
+    length : :obj:`float`
+        Shoreline length, computed as described above.
+
+    line : :obj:`np.ndarray`
+        If :obj:`return_line` is `True`, the shoreline, as an ``Nx2`` array of
+        x-y points, is returned.
```
elbeejay: Return statement needs to be updated for this function. I actually find the return behavior to be a less than intuitive - default case when `return_distances` is False, returns 2 numbers, the average distance to shoreline and standard deviation of distance to shoreline. But if `return_distances` is True, then just 1 object is output, the full set of distance values is returned. 

I would expect the function to always return mean / standard deviation, and if `return_distances` is True, then *also* return the set of distance values (so return 3 things in that case, mean, standard deviation, set of distance values).

But I guess this is a pretty simple thing and so I don't have a big problem either way, once the doc-string is revised it will be easy to use the function in practice.

elbeejay (approved):
With the exception of the return portion of the doc-string this looks good to me. 

I fiddled with using `scipy.ndimage.distance_transform_edt()` to calculate distances instead of the method implemented here but my alternative implementations were consistently slower.

deltametrics/plan.py
amoodie: Thanks for catching the docstring :roll_eyes: oops

Yeah, I think you're right. At first I was thinking that if someone wanted the distances, they would want them to do a histogram or their own percentiles, and would just compute the mean+-std themselves, and so wouldn't need the mean and std returned. 

But you're right, it is more intuitive to just return an additional argument. This is also consistent with the other `plan` functions then so that makes more sense.

Changed in the latest push.

PR #61, opened 2021-05-08
io-connection and doc-strings

### LINKED ISSUE #60: New xarray version has breaking changes for opening files that need to be addressed
New version of xarray 0.18 (http://xarray.pydata.org/en/stable/whats-new.html#v0-18-0-6-may-2021) has breaking changes involving backends used to open files. 

This seems to have created an issue with opening the Landsat.hdf5 example dataset, and is causing tests to fail on newer python versions that use 0.18 by default.

A quick fix would be to pin xarray temporarily to 0.16, but I think it's probably an easy fix to make the backend work again.

elbeejay:
Closes #60 by checking the file extension in `io.connect()`. To check for the HDF5 files and load them using the `xarray` engine, we need to have the `h5netcdf` package as a dependency... 

Before opening this PR for review/merging I'm planning to commit some smaller stuff to expand on and clean up doc-strings...

### DIFF
--- a/deltametrics/io.py
+++ b/deltametrics/io.py
@@ -163,8 +163,15 @@ class NetCDFIO(BaseIO):
                 self.data_path, "w", format="NETCDF4")
             _tempdataset.close()
 
+        if os.path.splitext(self.data_path)[-1] == '.nc':
+            _engine = 'netcdf4'
+        elif os.path.splitext(self.data_path)[-1] == '.hdf5':
+            _engine = 'h5netcdf'
+        else:
+            TypeError('File format current unsupported by DeltaMetrics.')
+
         try:
-            _dataset = xr.open_dataset(self.data_path)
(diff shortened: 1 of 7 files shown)


amoodie (approved):
LGTM. Weird that xarray seems to have formerly handled the hdf5 file with no problems, but they decided to drop the engine to load the files be default?

elbeejay:
Tried to contribute a bit to the doc-strings/documentation as part of the PR too.

> Weird that xarray seems to have formerly handled the hdf5 file with no problems, but they decided to drop the engine to load the files be default?

Yeah definitely unexpected, I had the tests fail locally because `xarray` doesn't have a backend engine for hdf5 unless you have that hdf5 library available. Maybe it is more flexible this way as they can lean on whatever the engine is from that hdf5 library rather than having their own or using one with limited features from the netCDF4 package?

PR #62, opened 2021-07-01
Channel measures

amoodie:
Implement simple channel width and depth measurements. These metrics compute geometries of channels along a predefined section. For example, the widths:

image


elbeejay (approved):
This looks fine to me, didn't take a super in-depth look but the logic and implementation make sense I think.

:+1:

PR #63, opened 2021-07-09
Add Morphological Planform

elbeejay:
This PR accomplishes the following:

1. Add an alternative `MorphologicalPlanform` based on a mathematical morphology method similar to the one described in [Geleynse et al 2012](https://doi.org/10.1029/2012GL052845)
2. Make the masking classes and methods more flexible to handle both the new "MPM" (morphological planform method) and existing "OAM" methodologies -- with the default behavior falling back on the OAM

The new `MorphologicalPlanform` class has been created, and the masking operations (and related unit tests) have been modified to allow the existing unit tests to pass, ensuring that the pre-existing OAM functionality has not been altered.

Of note, the `angle_threshold` has been generalized to a `contour_threshold` as this is more generic. For the OAM we use an angle threshold, but for the new MPM, this threshold is a value between 0 and 1. Similarly, the contour picking used for the shoreline extraction is based on an attribute `composite_array` that is just an alias to another field (for the OAM it is an alias for `sea_angles`). This is to just create a general and consistent name for the masking functions to refer to. 

Added an example of working with the two planform objects to the documentation, and included the shoreline extraction and a quick comparison. The two methods look roughly similar, of course the value used to pick the shoreline contour does matter, as does the "max disk" parameter used to control the range of disk sizes used for the morphological closing.

image

I think the code could probably be made cleaner / more elegant, however this works and has some minimal documentation. At some point we probably need to revisit the entirety of the `deltametrics` codebase one file at a time to try and refactor things and add documentation anyway...

### DIFF
--- a/deltametrics/_version.py
+++ b/deltametrics/_version.py
@@ -3,4 +3,4 @@ def __version__():
     """
     Private version declaration, gets assigned to dm.__version__ during import
     """
-    return '0.2.0'
+    return '0.3.0'
--- a/deltametrics/cube.py
+++ b/deltametrics/cube.py
@@ -428,7 +428,8 @@ class BaseCube(abc.ABC):
                        vmax=self.varset[var].vmax)
         cb = plot.append_colorbar(im, ax)
         if colorbar_label:
-            _colorbar_label = self.varset[var].label if (colorbar_label is True) \
+            _colorbar_label = \
+                self.varset[var].label if (colorbar_label is True) \
                 else str(colorbar_label)  # use custom if passed
             cb.ax.set_ylabel(_colorbar_label, rotation=-90, va="bottom")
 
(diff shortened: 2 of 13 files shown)


amoodie (changes requested):
Hi @elbeejay, I didn't really dig into all the refactoring, but I looked at the implementation of the new planform, and I think I understand the way this cascades into changes in masking.

I think it makes sense, and if you are happy with it, I'm good for you to go ahead with it. 

BUT. This seems like a pretty big breaking change for masking. So, I suggest we either increment the minor version number (so go `0.3.0`), or try to make the transition here smoother and then go with a smaller patch designation (so go `0.2.1`).  If you want to go the latter route, I think the mininum would be to make it so that the old mask tests could pass without the changes. This would mean implementing `dummy properties like `angle_threshold` to pass on to the new props and raise a deprecation warning. You probably have a better sense than I do for how much work that would be. To be clear, I'm fine with just going ahead with a 0.3.

Either way, the version number def needs to be changed.

elbeejay:
Sorry I didn't have it in me to dig through the PR and establish deprecation pathways. Will keep that in mind in the future as it is would be a better practice when making breaking changes. 

Just pushed a commit to jump the version to 0.3.0

amoodie:
Understood. I think it is important to try at least to some extent to keep things backwards compatible, now that we are starting to publish on work using DeltaMetrics. For example, I need to pin my faulting-subsidence stuff to version 0.2 now, or it would not work.

Go ahead and merge if you're ready.

elbeejay:
> Understood. I think it is important to try at least to some extent to keep things backwards compatible, now that we are starting to publish on work using DeltaMetrics. For example, I need to pin my faulting-subsidence stuff to version 0.2 now, or it would not work.
> 
> Go ahead and merge if you're ready.

Shouldn't be horrible to make backward-compatible if you think it is worth doing. Let me know and I can add the deprecation hooks

PR #64, opened 2021-08-20
Utility Function: Model runtime from the log-file

elbeejay:
Small PR, adds a utility function to get the model runtime from a logfile.

Associated test is marked to x-fail as `pyDeltaRCM` is not a dependency of `DeltaMetrics`. It is a tiny function, not sure how useful it is in the broad sense, but at least for me, knowing how long a run took to complete is nice to know when designing/planning a set of runs. Wasn't sure if this was a better fit here or in the `pyDeltaRCM` model repository itself.

### DIFF
--- a/deltametrics/utils.py
+++ b/deltametrics/utils.py
@@ -1,6 +1,8 @@
 import numpy as np
 import xarray as xr
 from scipy import optimize
+import time
+import datetime
 
 from numba import njit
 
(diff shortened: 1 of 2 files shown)


amoodie (approved):
I'm fine with this, if you find it helpful. Is "seconds" the unit you want for return though? Doesn't matter to me since it's just a utility function, so whatever unit is most helpful to you should be the return (hours?).

More broadly though, I'd be supportive of adding this info as output from pyDeltaRCM. Could be calculated at the end of the run during finalize and written to log? Maybe there is some way to write it into the netCDF file? Could open an issue on that repo about it. :+1:

elbeejay:
> More broadly though, I'd be supportive of adding this info as output from pyDeltaRCM. Could be calculated at the end of the run during finalize and written to log? Maybe there is some way to write it into the netCDF file? Could open an issue on that repo about it.

Not sure why this hadn't occurred to me, but I'll look and see if there's a clean way to add this information to the model directly. If so I don't see a need to merge this PR. Thanks!

amoodie:
Okay, feel free to merge this too though, could be helpful for legacy runs if you do add to pyDeltaRCM too

PR #66, opened 2021-09-22
Channelwidth bugfix

### LINKED ISSUE #65: Negative Channel Widths
Need to do more to document this and get to the root cause, but wanted to flag the issue. It appears as if the `compute_channel_width` function can return negative width values. 

The image below shows a binary channel map with a circular section shown in red. Computing channel widths along this section returns the following list of channel widths: 

```
array([-230.96551211,    5.        ,    4.        ,    3.41421356,
          2.41421356,    2.41421356,    1.41421356,    5.82842712,
          3.41421356,    3.        ])
```

image

We know channels cannot have negative widths, so therefore this is a bug.

elbeejay:
Aims to resolve #65. Does a bit of re-factoring to reduce the amount of repeated code for the width and depth functions. 

Basically looks for a 0-index value in the channel-start list and replaces it with a 1 if there is a 0. This means the channel width at the boundary will be underestimated by 1 pixel value, but this does prevent the negative width values from appearing... so maybe more of a temporary solution than a permanent fix here. 

EDIT: looks like this needs to go after #63 as I think I branched from there

### DIFF
--- a/deltametrics/plan.py
+++ b/deltametrics/plan.py
@@ -1261,16 +1261,11 @@ def compute_channel_width(channelmask, section=None, return_widths=False):
             'Input for `channelmask` was wrong type: {}.'.format(
                 type(channelmask)))
 
-    _channelseries = channelmask[section_trace[:, 1],
-                                 section_trace[:, 0]].astype(int)
+    # get channel stars and ends
+    _channelstarts, _channelends = \
+        _get_channel_starts_and_ends(channelmask, section_trace)
 
-    # compute the metrics
-    _padchannelseries = np.pad(_channelseries, (1,), 'constant',
-                               constant_values=(False)).astype(int)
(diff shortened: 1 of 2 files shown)


amoodie (approved):
is the only change here the last commit on this branch? If so, then it looks good. Maybe could just elaborate on the docstring for the internal helper (what it does more specifically and how it should be used) so we remember it exists and to use it.

elbeejay:
> is the only change here the last commit on this branch? If so, then it looks good. Maybe could just elaborate on the docstring for the internal helper (what it does more specifically and how it should be used) so we remember it exists and to use it.

Yes it's just the last commit, I'll try to bring the branch up to speed with 'develop' and hopefully that cleans up the list of old commits. Will expand on the doc-string :+1:

amoodie:
sounds good, feel free to merge when ready

PR #67, opened 2021-09-28
Optional return of image for dm.plot.aerial_view()

elbeejay:
Minor change but a useful one I think. Was trying to make frames for an animation and wanted to use `dm.plot.aerial_view()`. It is faster to update the image drawn by `plt.imshow()` than it is to clear the axes and re-draw it for each frame. To do this the `aerial_view()` function needs to return the image it collects when calling `plt.imshow()` internally. This PR adds this as an optional return which can be obtained by toggling the optional keyword argument `return_im` to `True`. 

Because this addition is an optional one, existing behavior of the function is unchanged.

Rough example of my use-case:
```
for i in range(timesteps):

    # conditional for 1st timestep
    if i == 0:
        _, im = dm.plot.aerial_view(cube['eta'][i, ...], return_im=True)

    else:
        im.set_data(cube['eta'][i, ...])
        plt.draw()


    # save frame
    plt.savefig(...)
```

### DIFF
--- a/deltametrics/plot.py
+++ b/deltametrics/plot.py
@@ -1,7 +1,6 @@
 import numpy as np
 
 import colorsys
-import copy
 
 import matplotlib
 import matplotlib.pyplot as plt
--- a/deltametrics/plot.py
+++ b/deltametrics/plot.py
@@ -1,6 +1,7 @@
 import numpy as np
 
 import colorsys
+import copy
 
 import matplotlib
 import matplotlib.pyplot as plt
@@ -1093,7 +1094,7 @@ def show_histograms(*args, sets=None, ax=None, **kwargs):
 
 
 def aerial_view(elevation_data, datum=0, ax=None, ticks=False,
-                colorbar_kw={}, **kwargs):
+                colorbar_kw={}, return_im=False, **kwargs):
     """Show an aerial plot of an elevation dataset.
 
     See also: implementation wrapper for a cube.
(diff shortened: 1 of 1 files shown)


deltametrics/plot.py
```diff
@@ -1,6 +1,7 @@
 import numpy as np
 
 import colorsys
+import copy
```
amoodie: was this added by mistake?

deltametrics/plot.py
```diff
@@ -1118,6 +1119,9 @@ def aerial_view(elevation_data, datum=0, ax=None, ticks=False,
     colorbar_kw
         Dictionary of keyword args passed to :func:`append_colorbar`.
 
+    return_im : bool, optional
```
amoodie, line 1121: not disagreeing with the choice of `return_im` here, but we will want to be consistent through the package. I think `im` makes sense as a general name for the image plotted by things, just wanted to make sure you thought about it in this context. Only alternative that comes to mind would be `return_image`, but `im` may actually be better because it makes it clearer `im` is an object (not just an array or something) from `imshow`.

amoodie (approved):
LGTM

deltametrics/plot.py
elbeejay: I think the only other place we do this is for the widths/depths of channels which use `return_widths` and `return_depths` respectively. So I agree, here either `return_im` or `return_image` could work

amoodie: we have some [optional returns in other unrelated functions too](https://github.com/DeltaRCM/DeltaMetrics/blob/b82c3e46b2a2730aab44f9904734108ca77a549e/deltametrics/plan.py#L687) and [here too](https://github.com/DeltaRCM/DeltaMetrics/blob/b82c3e46b2a2730aab44f9904734108ca77a549e/deltametrics/plan.py#L882). These are more specific to what is returned though. 

I think it's fine as is, unless you have any unifying idea...

PR #70, opened 2021-10-15
Dip section

elbeejay:
Pretty straightforward, just establishes the dip section functionality in `section.py`. Almost totally based on the strike section code so there isn't much that is actually new in this PR.


amoodie (approved):
yeah looks good to me, without having actually tested it. 

I think we'll have to revise some of this with the incoming xarray changes, but it's good for now. Thanks for implementing it!

PR #72, opened 2021-11-04
Temporary fix for show_plan method

### LINKED ISSUE #71: Fix cube.show_plan()
Somewhere along the way we broke `cube.show_plan`, see the [mask plot](https://deltarcm.org/DeltaMetrics/guides/userguide-3.png) from our user-guide as an example.

elbeejay:
Slightly modifies the `show_plan()` method to set x/y limits using metadata if it is available, otherwise previous behavior is retained at this time. Closes #71.

Although will have to be re-visited if the method is moved to `plan.py` and as the dimensions/coordinate handling changes.

### DIFF
--- a/deltametrics/cube.py
+++ b/deltametrics/cube.py
@@ -439,8 +439,15 @@ class BaseCube(abc.ABC):
         if title:
             ax.set_title(str(title))
 
-        ax.set_xlim(self.x[0], self.x[-1])
-        ax.set_ylim(self.y[0], self.y[-1])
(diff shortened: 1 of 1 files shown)


amoodie (approved):
yeah this is fine as a quick fix. I think I implemented something with setting the `extent` argument in the `imshow` call to get it working correctly, but I need to get a bunch of things merged in still, sorry about that.

elbeejay:
bit wary about this docs error but I guess we'll see what happens

PR #74, opened 2021-11-09
Integrate xarray -- Part 1

amoodie:
This is an attempt to integrate the new ordering of `x` and `y` from the latest pyDeltaRCM outputs. 

The signature to _correctly_ create a cube from a pyDeltaRCM model output is now:
```
cube.DataCube(golf_path, coordinates={'x': 'y', 'y': 'x'})
```

omitting the coordinates option will return a cube, but the orientation of slicing will be switched. Although a bit clunky, this allows us to flexibly import a netcdf file with arbitrary coordinates and then handle everything internally with a consistent 'x' and 'y'. 

This PR does some parts of integrating the xarray into DeltaMetrics, namely fixing section slicing to use coordinate arrays.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -56,7 +56,7 @@ jobs:
       run: |
         python -m pip install --upgrade pip
         pip install -r requirements.txt
-        pip install sphinx==2.0.1
+        pip install sphinx
         sudo apt update -y && sudo apt install -y latexmk texlive-latex-recommended texlive-latex-extra texlive-fonts-recommended dvipng
     - name: Install DeltaMetrics
       run: |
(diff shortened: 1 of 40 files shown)


elbeejay:
> My feeling is this is kind of clunky, but it is the easiest way to break as little as possible, while increasing flexibility. Thoughts?

I agree. Although I do prefer this to having the default be the DeltaRCM coordinates as those are opposite to typical conventions.

amoodie:
Right, maybe we could add some "detection" to the I/O that would automatically apply the coordinate switch if it detects it is connected to a pyDeltaRCM output file, but that would be a feature for another day...

I'll continue on trying to get this PR passing tests and merged then, as it is some progress towards the xarray integration.

PR #80, opened 2021-11-17
Refactor section inputs

### LINKED ISSUE #33: Create a `DipSection` section type
Create a section type similar to `StrikeSection`, but in the dip direction, as `DipSection`. Should take `x=` as an argument and create a section perpendicular to the shoreline. 

Follow the implementation of the `StrikeSection`.

Write tests.

### LINKED ISSUE #34: Make the `Section` types work with the coordinates of the arrays from xarray
for example, when creating a `StrikeSection`, `y=10` should be at the actual coordinate `y==10`, not the 10th cell in the `y` direction. This will be a breaking change, but I feel it is the correct way to handle this as input. 

One option would be to implement another option `y_cell=10` that is mutually exclusive with `y=`, and maintains the functionality of the current api.

amoodie:
Refactored the inputs to sections (closes #34)

- [x] PathSection
- [x] StrikeSection
- [x] DipSection
- [x] CircularSection
- [x] RadialSection

Each section can take inputs as either dimensional coordinates or as indices. Each argument has a paired argument `xxxx_idx` that is the input variables. Documentation and tests updated,

**warning: this is a breaking change. The behavior of some input arguments has been maintained with a deprecation warning (e.g., `y` as input for a strike section), but others could not maintain backwards compatability (e.g., `radius` for a circular section now specified coordinates instead of indices and will likely result in an error without updating your own code).

Note: this PR also fixed a few issues with sections. Namely, `Path` and `Radial` sections would occasionally be drawn the wrong direction or have incorrect ordering.

Will also complete and closes #33

### DIFF
--- a/deltametrics/section.py
+++ b/deltametrics/section.py
@@ -665,7 +665,89 @@ class PathSection(BaseSection):
         return self.trace
 
 
-class StrikeSection(BaseSection):
+class LineSection(BaseSection):
+
+    def __init__(self, direction, *args, distance=None, distance_idx=None, length=None,
+                 x=None, y=None, **kwargs):
+        """Initialization for the LineSection.
+
+        The LineSection is the base class for Strike and Dip sections,
+        as these share identical input arguments, and processing steps, but
+        differ in which direction the line is drawn.
+
+        .. note:: the `RadialSection` does not subclass `LineSection`.
+        """
+
+        self._distance = None
+        self._distance_idx = None
+        self._length = None
+        self._length_idx = None
+
+        # process the optional/deprecated input arguments
+        #   if y or x is given, cannot also give distance idx or length
+        if ((not (y is None)) or (not (x is None))):
+            #   check if new args are given
+            if (not (distance is None)) or (not (distance_idx is None)) or (not (length is None)):  # noqa: E501
+                raise ValueError(
+                    'Cannot specify `distance`, `distance_idx`, or `length` '
+                    'if specifying `y` or `x`.')
+            #   if new args not given, then use old args in place of new
+            else:
+                warnings.warn(
+                    'Arguments `y` and `x` are deprecated and will be removed'
+                    'in a future release. Please use `distance_idx` and '
+                    '`length` to continue to specify cell indices, or '
+                    'use `distance` and `length` to specify '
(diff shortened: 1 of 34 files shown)


amoodie:
@elbeejay can you build the docs and read through the changes to the `Section` API description? Just want to make sure you agree with all the names and changes before merging. Don't expect you to dig into the code, unless you want to.

deltametrics/section.py
```diff
@@ -800,104 +1078,182 @@ class DipSection(BaseSection):
...
 class CircularSection(BaseSection):
     """Circular section object.
 
     Section drawn as a circular cut, located a along the arc a specified
-    radius from specified origin.  Specify the location of the circular section
-    with :obj`radius` and :obj:`origin` keyword parameter options. The
-    circular section trace is interpolated to the nearest integer model domain
-    cells, following the mid-point circle algorithm
+    `radius` from specified `origin`.  Specify the location of the circular
+    section with :obj`radius` and :obj:`origin` keyword parameter options.
```
elbeejay: missing a colon in reference to radius object here

deltametrics/section.py
```diff
@@ -912,57 +1268,146 @@ class CircularSection(BaseSection):
         section, or the `Cube` is passed as the first positional argument
         during instantiation.
 
+    .. warning::
```
elbeejay: this looks right, but didn't render as expected 
image

deltametrics/section.py
```diff
@@ -974,10 +1419,17 @@ class RadialSection(BaseSection):
     nearest integer model domain cells, following the a line-walking algorithm
```
elbeejay, line 1419: Old line above, reference to azimuth object is missing a colon

elbeejay:
> @elbeejay can you build the docs and read through the changes to the `Section` API description? Just want to make sure you agree with all the names and changes before merging. Don't expect you to dig into the code, unless you want to.

Just read through the docs and they look really good. The names and everything feel very consistent and are intuitive I think.

amoodie:
@elbeejay sweet, thank you. The documentation is so fragile to formatting :weary: 

reminder, this will be a breaking change if you pull after merge!

PR #83, opened 2021-12-03
New data: aeolian dune field

amoodie:
add some new data from a different domain.

This data is from an aeolian dune field numerical model.

image

The main goal with this contribution is to have a dataset to test and develop with that is different from the pyDeltaRCM outputs.
This dataset:
* uses different names for the coordinates (easting and northing)
* does not have anything in the `meta` field
* is a dune field, so it's not a delta at all

hoping this helps move us towards being more flexible for domains other than deltas (and then change the name 😀)

NOTE: This did help me find a bug with the way we pulled coordinate names, already! So that's cool, and I fixed it.

### DIFF
--- a/deltametrics/io.py
+++ b/deltametrics/io.py
@@ -1,6 +1,7 @@
 
 import abc
 import os
+import copy
 from warnings import warn
 
 import xarray as xr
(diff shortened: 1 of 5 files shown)


elbeejay:
really cool, good idea :+1:

PR #88, opened 2022-01-07
add a DepositMask

amoodie:
add a basic deposit mask function for quickly determining where sediment has been deposited, and having the benefits of type checking and handling  mask-dimension and plotting.

Example: 
image

### DIFF
--- a/deltametrics/mask.py
+++ b/deltametrics/mask.py
@@ -2374,14 +2374,15 @@ class DepositMask(BaseMask):
     .. plot::
         :include-source:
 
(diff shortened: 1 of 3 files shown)


elbeejay (approved):
This looks good to me :+1:

I agree that there's potential to link this new DepositMask and the ThresholdValueMask. Might be worth dropping an issue just to keep that on our radar before someone jumps in and implements all the pathways for the DepositMask.

amoodie:
"someone" 🙄

Issue #87, opened 2022-01-07
Update installation instructions


elbeejay:
In the [developers' installation guide](https://deltarcm.org/DeltaMetrics/guides/devguide.html#installation) we currently skip stating that the dependencies (available in the *requirements.txt* file) need to be installed before trying to install the editable version of the package.

So basically this line:
```
pip install -r requirements.txt
```
should be added after changing into the cloned directory and before doing the editable install.

elbeejay:
Should have been closed with #89

PR #89, opened 2022-01-07
update docs, close issues

### LINKED ISSUE #68: Expand internal function (_get_channel_starts_and_ends) doc-string
Forgot to do this before merging #66... will try to fold it into the next PR that improves documentation.

> Maybe could just elaborate on the docstring for the internal helper (what it does more specifically and how it should be used) so we remember it exists and to use it.

elbeejay:
Closes #68 and #87.


amoodie (approved):

elbeejay:
going to merge this now as it includes adjustments to the `pytest.approx()` calls in the tests which will break CI builds for other branches soon enough

PR #90, opened 2022-01-12
Output dimension renaming utility

elbeejay:
I've managed to need to analyze outputs from pyDeltaRCM versions before and after our breaking change to the dimensions in the output file (v2.1.0). But beyond my own needs for messing with the dimensions of netcdf files, this utility function might have value when coming from another model and trying to use our analysis package; if the dimensions are not time/x/y they can be easily re-named.

Edit: although is this widely useful? I don't know. I'm going to use it now but can't imagine I use it in 6 months for example. So I'm indifferent on it merging.


elbeejay:
Commands for renaming dimensions and variables in netCDF files are so trivial `renameDimension` and `renameVariable` that I've decided against adding this to the codebase. :-1:

amoodie:
Okay. I started working on, but haven't finished, a guide on how to set up the netcdf file to work correctly with DM. I think the code in here could be included as a "Renaming variables in an existing netCDF file" example in the guide :+1:

PR #98, opened 2022-02-25
close issue #97

### LINKED ISSUE #97: future warning on morphological operations
```
/opt/hostedtoolcache/Python/3.8.12/x64/lib/python3.8/site-packages/DeltaMetrics-0.3.0-py3.8.egg/deltametrics/plan.py:1635: FutureWarning: `selem` is a deprecated argument name for `binary_closing`. It will be removed in version 1.0. Please use `footprint` instead.
```

elbeejay:
replaces the 'selem' kwarg for the scikit-image binary closing method with the new
'footprint' argument. sets the minimum version for scikit-image to 0.19 in the
requirements.txt file which is the first version that has the 'footprint' argument.

closes #97

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -11,7 +11,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        python-version: [3.6, 3.7, 3.8]
+        python-version: [3.7, 3.8, 3.9]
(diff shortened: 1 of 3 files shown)


elbeejay:
So this is interesting; making this change would drop support for Python 3.6 -- are we okay with that @amoodie? if so I can add a commit to drop the 3.6 CI builds.

PR #104, opened 2022-03-07
Fix eg

### LINKED ISSUE #103: broken doc example
The MPM example with the disk is broken, because the max size disk is too small. 

https://deltarcm.org/DeltaMetrics/guides/examples/planforms/planform_intro.html

amoodie:
Fix the example broken in #103, easy fix. Just needs a value to better reflect the width of the mouth.

before:
image

after:
image

One option might be to add a check in the [mask creating bits here](https://github.com/DeltaRCM/DeltaMetrics/blob/a386f0b3b98ff61ea976e33ac2842148094cc173/deltametrics/mask.py#L1479) to check if there are multiple contours returned for the threshold selected? We could at least raise a warning that the mask might not be as expected because two contours were identified (and suggest increasing the max_disk value). 

Related: might be helpful to raise a warning if the `contour_threshold` is greater than any values in the field being contoured. This would be helpful here to say MPM field is between 0 and 1, so select a threshold in there?

Thoughts?

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -8,12 +8,13 @@ DeltaMetrics
 .. image:: https://codecov.io/gh/DeltaRCM/DeltaMetrics/branch/develop/graph/badge.svg
   :target: https://codecov.io/gh/DeltaRCM/DeltaMetrics
 
-DeltaMetrics is a Python package for manipulating depositional system data cubes.
-In particular, DeltaMetrics has robust objects and routines designed to help organize, visualize, and analyze topography and sedimentological timeseries data deltaic systems (e.g., the pyDeltaRCM numerical model and laboratory delta experiments).
+*DeltaMetrics* is a Python package for manipulating depositional system data cubes.
+The package has robust objects and routines designed to help organize, visualize, and analyze topography and sedimentological timeseries data.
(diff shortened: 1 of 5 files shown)


elbeejay:
> One option might be to add a check in the [mask creating bits here](https://github.com/DeltaRCM/DeltaMetrics/blob/a386f0b3b98ff61ea976e33ac2842148094cc173/deltametrics/mask.py#L1479) to check if there are multiple contours returned for the threshold selected? We could at least raise a warning that the mask might not be as expected because two contours were identified (and suggest increasing the max_disk value).

I think the warning makes sense - not sure how to tie in the suggestion as that contour function is common to both the OAM and MPM methods.

> Related: might be helpful to raise a warning if the contour_threshold is greater than any values in the field being contoured. This would be helpful here to say MPM field is between 0 and 1, so select a threshold in there?

Yeah, I don't know if this type of input checking is something we should just generally have more of? What if the threshold is negative? etc etc.

amoodie:
>Yeah, I don't know if this type of input checking is something we should just generally have more of? What if the threshold is negative? etc etc.

I think wherever you can reasonably check on values, it's pretty helpful. It usually costs next to nothing in computation, but provides a lot of helpful context when writing code.


I've opened an issue #105 to track this issue with the MPM threshold, so I'm going to merge this fix for now.

PR #106, opened 2022-03-10
Computing stratigraphy in presence of subsidence

elbeejay:
This PR enables us to account for subsidence when computing stratigraphy.

The real addition here is the private function `strat._adjust_elevation_by_subsidence` which has been designed in a similar vein to the other private stratigraphy functions (e.g., `_compute_preservation_to_cube` or `_determine_strat_coordinates`). Subsidence is assumed to be specified as a *rate* (`sigma`) at some time, as opposed to a cumulative distance at a given time. Optional `sigma` arguments have been added to the higher-level stratigraphy functions which lead to calls of this private function when subsidence rate information is provided; otherwise existing behavior is maintained by default.

Design-wise, we could alter the higher-level calls to set `sigma=0` as the default value, and **always** call the new function - I haven't tested to see whether this would be faster or slower than the if-statement used now (although I suspect it would be slower). There is also the option to separate the type / shape handling in the new function from the actual elevation adjustment, I wasn't sure exactly where this would best fit if not in the function itself, and it feels similar enough to me, to the functionality in `_determine_strat_coordinates` as currently arranged.


elbeejay:
After a conversation w/ @amoodie we decided some changes need to be made to this PR:

- [x] instead of providing a *rate* `sigma` expect input to be a "distance subsided" at a given time
- [ ] <s>can still have some basic behavior for `sigma` but need to properly define assumptions, boundaries, and limitations of its use</s>
- [ ] add some more exotic test cases to capture edge behavior for stratigraphic preservation when there is subsidence happening
- [x] double-check that final implementation is utilized in all higher-level functions and the subsidence parameter gets passed through properly
- [x] increment minor version of package

so i'll slowly get around to implementing this stuff

elbeejay:
After a conversation w/ @amoodie we decided some changes need to be made to this PR:

- [ ] instead of providing a *rate* `sigma` expect input to be a "distance subsided" at a given time
- [ ] can still have some basic behavior for `sigma` but need to properly define assumptions, boundaries, and limitations of its use
- [ ] add some more exotic test cases to capture edge behavior for stratigraphic preservation when there is subsidence happening
- [ ] double-check that final implementation is utilized in all higher-level functions and the subsidence parameter gets passed through properly

so i'll slowly get around to implementing this stuff

elbeejay:
>  add some more exotic test cases to capture edge behavior for stratigraphic preservation when there is subsidence happening

Don't remember what "exotic" conditions I'd been thinking of when I wrote this item. I did check the unit tests I wrote for the subsidence functionality against 1-D sketches I'd drawn, so the function is working as I expect it to. Included a case with some uplift, to make sure the function is agnostic to the direction of the vertical change.

So I think this is ready to go? Ended up dropping support for the "rate" `sigma` in favor of the simpler `sigma_dist` which is the cumulative distance subsided at any time.

PR #116, opened 2022-06-24
add style kilometer tick formatter and axes formatter function.

amoodie:
A simple utility function for plotting, that changes axes tick labels from meters to kilometers. I constantly find myself wanting this as a function, and so copying and pasting it from one notebook to the next or googling to recreate it. Still, I'm not sure if it's too specific to my needs to belong in the package...

image

Todo:
- [x] tests


elbeejay:
Looks handy, I would certainly use it if merged

PR #117, opened 2022-07-20
update deploy yaml

elbeejay:
updates the deploy.yml file to match what we do in pyDeltaRCM.

zenodo integration has been set-up and toggled on so the next release should automatically archive on zenodo and get a DOI

@amoodie the DeltaMetrics repository just needs the pypi secrets before we trigger a new release and have this workflow run (maybe we do that before merging this and use this PR to increment the version too...)

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -17,11 +17,12 @@ jobs:
     - name: Set up Python
       uses: actions/setup-python@v2
       with:
-        python-version: '3.x'
+        python-version: '3.10'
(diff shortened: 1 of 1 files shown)


amoodie:
looks good, thanks for doing this. 

I just added the secrets, so it _shoul_ deploy to pypi on a github tagged release. Looks like our last release was 0.2.0, and now we are at 0.3.2, so incrementing isn't necessary, I don't think.

amoodie (approved):

PR #120, opened 2022-08-25
Add some data from Savi et al., 2020 and the channelized response variance (CRV) metric

elbeejay:
This PR does 2 things:

1. Adds sample data (topographic scans and overhead imagery) from one of the experiments conducted in [Savi et al., 2020](https://doi.org/10.5194/esurf-8-303-2020) (reformatted and stored as a zipped pair of netCDF files as a [zenodo dataset](https://doi.org/10.5281/zenodo.7047109)).
2. Adds the channelized response variance (CRV) metric

The overhead images from the new Savi et al., 2020 sample dataset are used to generate the following example in the `calculate_channelized_response_variance()` docstring:

image


amoodie (commented):
# summary
I didn't check the details of the CRV implementation, but it seems reasonable to me. I just have some comments on code organization and sample data.

## Sample data

I checked out the sample data, and the cube seems to work correctly, but I think the coordinates did not come through? E.g., time in the description says that it should be every 30 minutes, but the coordinates just come through as sequential integers. The same is true of the x and y coordinates. 

I think we need to get this correct in the sample data. This could be a helpful example for the experimental community to see how their data could/should be organized, so I think it is important to get it right.

In the description of the sample data, should we make the overhead images display as RGB composited (e.g., follow the code in the sample `landsat` data)?

## Organization

Keeping with the rest of the package, I think we should rename the crv and related functions, and perhaps reduce the number of functions. It's not clear to me that all of these steps need to be separate functions, unless they have some utility beyond the CRV (e.g., the slopes function).

* `calculate_crv` --> `calculate_channelized_response_variance`
* `calculate_crv_mag` --> does this need to be public? Does it have any utlity outside the crv calculation? I suggest we make it either `_calculate_temporal_variance` or something, or just fold it back into the crv function?
* `calculate_slopes` --> this could be useful for more than the crv? Maybe this remains as a separate function, but private? `_calculate_temporal_linear_slope`?
* `calculate_directional_crv` --> maybe this should just be folded into one function (`calculate_channelized_response_variance`) with a flag as an option `directional=True`, as default?

amoodie:
After looking at the data again, I think it is actually correct. I think I was confusing the time interval and resolution of the images versus the scans. Sorry for the confusion!

elbeejay:
Thanks for the feedback @amoodie!

### re: sample data

While the units were correct in the netCDF, the dimensions actually were not. I've flipped the data so that the inlet is now at the top and flow moves downward (this is consistent with how our pyDeltaRCM outputs end up in deltametrics). The version of the dataset has been incremented to v1.1 and the registry was updated to reflect the correct hash and path.

### re: organization

Your suggestions were well taken, many of the functions were just a few lines. I've consolidated the variance calculation and slope thresholding/direction parts into the single `calculate_channelized_response_variance` function, and made the slope calculation function private, with a more appropriate name as suggested (now `_calculate_temporal_linear_slope`).

amoodie (approved):

PR #129, opened 2022-12-03
codecov-job

elbeejay:
only push coverage results to codecov for one job, not all the different versions

right now we get a bunch of failing tests because coverage results couldn't be pushed to codecov, this seems unnecessary, we can just have one job send results to codecoverage, the others can just install and run the tests.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -36,6 +36,26 @@ jobs:
     - name: Test with pytest
       run: |
         pytest --cov=deltametrics/ --cov-report=xml
+        
+  codecov:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python 3.9
+      uses: actions/setup-python@v2
+      with:
+        python-version: '3.9'
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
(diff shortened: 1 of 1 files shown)


amoodie:
Yeah, this seems reasonable. 

I think this setup was probably borrowed form pyDeltaRCM, where we also test and do coverage on different OS. I think there it makes a bit more sense, since we do have different behaviors for different OS, including some code blocks that are only reachable on the different OS. So, to get coverage you have to test and upload on the different OS.

As far as I can remember, we don't have any OS-specific behaviors like that in DeltaMetrics. Still though, we do want to test on the different OS. Uploading the coverage from only one OS might be fine.

elbeejay:
The tests are still being run for all cases. This PR proposes that we only push to codecov for one ubuntu case as that connection seems to be fragile in some instances (unless I've been misinterpreting the failed tests recently)

amoodie:
maybe a more flexible "fix" would be to just not care if the codecov upload fails?

[i.e., set this to false?](https://github.com/DeltaRCM/DeltaMetrics/blob/28c1d64eec76c764d48e00b356b570ced3af6eea/.github/workflows/build.yml#L46)

elbeejay:
I'm happy with that. Anything to reduce the number of "build failing" emails I get every month :smiling_face_with_tear:

PR #134, opened 2023-02-03
fix wetmask from planform bug

### LINKED ISSUE #93: Incomplete / Bug in WetMask.from_Planform pathway
If you try to instantiate a `WetMask` using the `WetMask.from_Planform` pathway it looks for an attribute `below_mask` in the `Planform` object that is passed. This attribute only exists for the OAM type of planform. So we need to either require `below_mask` for all `Planform` objects (via addition to the abstract class), or change the way the `WetMask.from_Planform` pathway works, or just throw a warning if the input planform doesn't have that attribute. 

Same issue is present for the same pathway for the `EdgeMask` as well as the `EdgeMask.from_Planform_and_WetMask` pathway.

elbeejay:
fixes #93 by adding the `below_mask` attribute to the `MorphologicalPlanform` class

### DIFF
--- a/deltametrics/plan.py
+++ b/deltametrics/plan.py
@@ -1040,7 +1040,7 @@ class MorphologicalPlanform(SpecialtyPlanform):
         self._shape = self._elevation_mask.shape
 
         # assign below mask
-        self._below_mask = ~(self._elevation_mask)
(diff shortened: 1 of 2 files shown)


amoodie (approved):
lgtm :+1:

PR #137, opened 2023-02-10
Compensation metric

amoodie:
Adds a function to compute compensation.

todo
 - [ ] ~rework how `strata` is pulled for a `Section`. `sect.strata` should return a 2d DataArray with strata for that section.~
 - [ ] ~clarify and streamline `Cube.strata`, and how `Section` gets this information.~
 - [x] documentation example for how to use do the computation and briefly what to do with the outputs. E.g., how to do a bin average, how to fit a line to the bin averages, what the rollover point means. Use the equations. 
 - [x] tests (API only, not actual code results)

### DIFF
--- a/docs/source/guides/examples/computations/stratigraphy_temporal_subset.rst
+++ b/docs/source/guides/examples/computations/stratigraphy_temporal_subset.rst
@@ -1,0 +1,9 @@
+Temporal subsets of stratigraphy
+--------------------------------
+
+
+TODO...
+
+
+* demonstration of computing stratigraphy for a subset of time of a dataset
+* make a movie of the strat evolving through time
--- a/docs/source/guides/examples/computations/stratigraphy_temporal_subset.rst
+++ b/docs/source/guides/examples/computations/stratigraphy_temporal_subset.rst
@@ -1,9 +1,108 @@
 Temporal subsets of stratigraphy
 --------------------------------
 
+This example creates straigraphy for a temporal subset of the dataset.
+This allows for a view of *time evolution of stratigraphy itself*, for example, the changing stack of an experiment through time.
 
-TODO...
+In this example, we make the subsets into a movie, but the relevant lines for creating a subset are:
+
+.. code::
+
+    vol, elev = spl.strat.compute_boxy_stratigraphy_volume(
+        aeolian_eta[:i],
+        aeolian_time[:i],
+        z=zs,
+        sigma_dist=subsidence_rate,
+    )
+
+where the `:i` indicates the slice of all of the times to be used in the calculations. See :obj:`~sandplover.strat.compute_boxy_stratigraphy_volume`.
+
+.. plot::
+    :include-source:
+    :context: reset
+    :nofigs:
+
+    import numpy as np
+    import matplotlib.pyplot as plt
+    from matplotlib import animation
+
+    import sandplover as spl
+
+    # set up to use constant z for stratigraphy, so all frames line up over time
+    zs = np.arange(-10, 10, 0.25)
+
+    # load the sample data and define indices to use in the plotting
+    aeolian = spl.sample_data.aeolian()
+    time_idxs = np.arange(100, 300, step=10)
+
+    aeolian_eta = aeolian["eta"].copy()
+    aeolian_time = aeolian["time"].copy()
+
+    subsidence_rate = 0.01
+    cross_section_idx = 10
 
 
-* demonstration of computing stratigraphy for a subset of time of a dataset
-* make a movie of the strat evolving through time
+    def update_field(i):
+        vol, elev = spl.strat.compute_boxy_stratigraphy_volume(
+            aeolian_eta[:i],
+            aeolian_time[:i],
+            z=zs,
+            sigma_dist=subsidence_rate,
+        )
+        im.set_data(vol[:, :, cross_section_idx])
+
+
+    # make one strat for the initial, to have the full shape of the image
+    vol, elev = spl.strat.compute_boxy_stratigraphy_volume(
+        aeolian_eta[: time_idxs[0]],
+        aeolian_time[: time_idxs[0]],
+        z=zs,
+        sigma_dist=subsidence_rate,
+    )
+
+    # make the first frame
+    fig, ax = plt.subplots(figsize=(6, 3))
+    im = plt.imshow(
+        vol[:, :, cross_section_idx],
+        extent=[0, aeolian.dim1_coords[-1], elev.min(), elev.max()],
+        aspect="auto",
+        origin="lower",
+        vmin=0,
+        vmax=time_idxs[-1],
+    )
+
+Then, make the animation.
+
+
+.. plot::
+    :context:
+    :nofigs:
+
+    import os
+
+    os.makedirs("../../../../build/plot_directive/guides/examples/plot/", exist_ok=True)
+
+    anim = animation.FuncAnimation(fig, update_field, frames=time_idxs, interval=5)
+    anim.save('../../../../build/plot_directive/guides/examples/plot/strat_movie.gif', fps=5)
+
+    anim = animation.FuncAnimation(
+        fig, update_field,
+        frames=time_idxs,
+        interval=5,
+        blit=False)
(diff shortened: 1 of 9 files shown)


amoodie:
@ericbarefoot I think this is ready for your review now. I would appreciate some review of the description and calculation of the compensation values. There is an example that shows how it works.

I still need to add tests.

amoodie:
Now has tests for the compensation function API, but actual calculation results. I'm not sure how to make a simple dataset to test.

docs/source/guides/examples/computations/stratigraphy_temporal_subset.rst
ericbarefoot, line 1: This one is very cool.

sandplover/section.py
```diff
@@ -395,6 +395,33 @@ def variables(self):
...
+                name=var,
+                attrs={
+                    "slicetype": "data_section",
+                    "knows_stratigraphy": self._underlying._knows_stratigraphy,
+                    "knows_spacetime": True,
+                },
+            )
+            return _xrDA
+        else:
+            raise utils.NoStratigraphyError(obj=self, var="strata")
```
ericbarefoot, line 423: This needs to be just a `NoStratigraphyError()` call since you import from `utils` at the top of the script.

sandplover/section.py
```diff
@@ -395,6 +395,33 @@ def variables(self):
         """List of variables."""
         return self._variables
 
+    @property
+    def strata(self):
+        """Stratigraphic surfaces array.
+
+        See :obj:`~deltametrics.cube.DataCube.strata` for more information.
```
ericbarefoot, line 402: I think you mean `sandplover.cube.StratigraphyCube.strata`

sandplover/section.py
```diff
@@ -395,6 +395,33 @@ def variables(self):
...
+        ------
+        NoStratigraphyError
+            If no stratigraphy information is found for the section.
+        """
+        if self._underlying._knows_stratigraphy:
+            _xrDA = xr.DataArray(
+                self._underlying.strata[:, self._dim1_idx, self._dim2_idx],
+                coords={"s": self._s, self._z.dims[0]: self._z},
+                dims=[self._z.dims[0], "s"],
+                name=var,
```
ericbarefoot, line 414: I think this `var` has no reference. What's it's scope? I thin you are looking to pass the variable, but I don't know the best way to do that right now.

sandplover/section.py
ericbarefoot, line 1: Not sure if I am quite understanding this property of section, and how the call works. It seems to function, but looks like there's a few bugs.

sandplover/cube.py
```diff
@@ -864,6 +864,15 @@ def from_DataCube(
             elevation data. Typically, this is ``'eta'`` in pyDeltaRCM model
             outputs.
 
+        sigma_dist : :obj:`float`, :obj:`list` of `float`, optional
+            Subsidence distance per timestep or list of subsidence per
+            timestep. When a singular (integer or float) value for subsidence
+            is provided, it is assumed that the provided value is the rate of
+            subsidence in terms of some vertical distance per timestep.
+            Conversely, when a time-series is provided, the each value is
```
ericbarefoot, line 872: typo: 

> Conversely, when a time-series is provided, ~the~ each value is

sandplover/cube.py
ericbarefoot, line 1: Can you clarify what the special value of this is? I am struggling to see where this gets used, and why subsidence is not just one of many auxiliary variables (or a spatially-explicit variable?). Why does it show up here? I think I'm just not understanding.

sandplover/strat.py
```diff
@@ -7,30 +7,201 @@
...
+
+    * :doc:`Example on Compensational Stacking <../guides/examples/computations/compensational_stacking>`
+
+    """
+    if time_idxs is None:
+        time_idxs = np.arange(stratal_surfaces.shape[0])
+    else:
+        raise NotImplementedError("Only works for processing all surfaces right now.")
+        # how do we handle a case where time intervals are known (rather than idxs)
+        #   and the intervals are not equally spaced?
```
ericbarefoot, line 99: great question. no idea

ericbarefoot: lemme think about it

sandplover/strat.py
```diff
@@ -7,30 +7,201 @@
...
+        dim0 = stratal_surfaces.dims[0]
+        times = stratal_surfaces[dim0][time_idxs]
+    else:
+        # use indexes = 1
+        times = np.arange(len(time_idxs), dtype=float)
+
+    if stratal_surfaces.ndim != 2:
+        raise ValueError("Not able to handle 3d or 1d yet.")
+        #  Note, that a one-dimensional array will always return a `sigma`
+        #  array of all ones. How to compute over 3d?
```
ericbarefoot, line 112: 3D is a reasonable extension of the 2D routine. 1D doesn't make a lot of sense, but we could consider it. would need multiple samples in 1D tho. is not defined for a single "core"

sandplover/strat.py
```diff
@@ -7,30 +7,201 @@
...
+    # prepare the data for the jitted (?) computation routine
+    if np.any(np.isnan(stratal_surfaces)):
+        # this could probably be relaxed, to clip out any
+        #    columns with any nans
+        raise ValueError(
+            "NaN found in stratal surfaces. Please remove all NaN values "
+            "before proceeding."
+        )
+
+    # run the computation routine (jitted?)
```
ericbarefoot, line 131: Do you want to enforce/require `numba`? That might not be a bad idea.

sandplover/strat.py
ericbarefoot, line 1: Implementation looks fine to me.

tests/strat_test.py
ericbarefoot, line 1: This definitely tests the error functions. Should we also have a test for making sure the function does what it says? Like a predetermined outcome on the golf dataset or something?

ericbarefoot (changes requested):
Mostly a few small changes that should be made. Core function is fine. Happy to make it a priority to put a test dataset on zenodo that `pooch` can fetch for actual compensation stats. Could coarsen it for ease of use.

sandplover/cube.py
amoodie: Subsidence can be just a regular value, but we need to know which of your many variables *is* subsidence if you want us to apply it during the calculation of preserved sediments. By default we assume zero. It could be spatially-explicit, but we don't currently support that (aka I didn't implement that yet).

sandplover/strat.py
elbeejay: One idea would be to resample yourself to the smallest time interval present in the data. Say you had time intervals of 1 year, 2 years, 5 years, with corresponding three indices, 0, 1, 2, you could resample to a bigger volume with equal spacings of 1 year (8 indices), and then I'd imagine your existing methods would work. Let's you maintain the highest temporal fidelity you have, with the assumption that you can somehow resample or subdivide the longer stretches of time into that smallest common interval (and at a higher computational cost).

sandplover/section.py
amoodie: thanks for noticing all the issues here. I think was something I never fully finished implementing. Now it has a better explanation and an example of what you can use this attribute for.

sandplover/strat.py
amoodie: opened issue #237, not addressing this in this PR

sandplover/strat.py
amoodie: sandplover already requires numba. I didn't end up using it for this function at present, but it could be changed in the future if this is a computation bottleneck

tests/strat_test.py
amoodie: I've tried to avoid predetermined calculations, as they seem difficult to maintain and understand (originally did this with some of the shoreline roughness calculations, for example). 

If there was a simple synthetic case we could use, where the answer is known, that would work. Like a random normal distribution of elevation changes plus background aggradation should equal compensation of xxxxx. 

I don't know enough to be able to implement this tho

amoodie: I opened an issue to track this testing as well #238

amoodie:
Thanks for the helpful review @ericbarefoot !

PR #138, opened 2023-02-26
Expand mobility documentation

elbeejay:
- [x] expand mobility guide
- [x] impove mobility function doc-strings


amoodie (approved):
This is awesome. It's much easier to understand what the different metrics are now (and where they actually come from originally). The expanded examples are great too. 

👍

PR #194, opened 2025-02-05
Rename package from deltametrics to sandplover

mcflugen:
We decided to rename the *deltametrics* package to be *sandplover*. This pull request, I hope, does that.

* The python package is now `sandplover`.
* In text, the package is referred to as *SandPlover*.
* The convention for importing `sandplover` is now `import sandplover as spl`.
* Links now point to github.com/sandpiper-toolchain

I think I got most stuff but I'm sure there are things that are missing.


mcflugen:
Note that [8aa56bf](https://github.com/sandpiper-toolchain/sandplover/pull/194/commits/8aa56bf7855d4b6264239b7b9cbb21cf16c00800) modfies the URL for the sample data to point to a file on the *develop* branch that still uses a path that contains the *deltametrics* folder. I had to make that change to get the tests to pass for this pr. Before merging, I will need to undo that commit so the path correctly points to the new *sandplover* folder.

amoodie (approved):
There's obviously too many changes here for me to check them all, but it looks in line with what we have discussed. 

- increment the version number, since this will almost definitely break codes using the "deltametrics" name. 
- I think we want to do just lowercase `sandplover` as the name for the project everywhere. We can make this change separately though, not a big deal.

I'm good to merge once the version is incremented.

mcflugen:
> There's obviously too many changes here for me to check them all

Yeah, sorry about that. I considered trying to break it up into smaller bits but any way you slice it, it was going to be a lot of small changes everywhere.

> I think we want to do just lowercase *sandplover* as the name for the project everywhere

Sounds good. Rather than do this in a separate pull request, I'll just remove the commits from this pr.

> increment the version number, since this will almost definitely break codes using the "deltametrics" name

Are you thinking version 1.0 or 0.5? If we're using semantic versioning, I think 0.5 should be fine since `0.x` versions allow for breaking changes.

amoodie:
No worries! I just meant I wasn't going to give a line-by-line review ha ha.

I think that we should still be in the 0.x territory. There are likely to be more breaking changes as we iterate the package with the sandpiper project/community.

mcflugen:
> I think that we should still be in the 0.x territory.

I agree. We're now at *sandplover* v0.5.0.

PR #198, opened 2025-02-20
Update the deploy workflow

mcflugen:
I've updated the *sandplover* workflow to use the [pypa/gh-action-pypi-publish](https://github.com/pypa/gh-action-pypi-publish) action to upload distributions to PyPI.

Instead of using `PYPI_USERNAME` and `PYPI_PASSWORD`, I've added our GitHub organization as a trusted publisher for *sandplover* (for both PyPI and Test-PyPI).

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -15,6 +15,16 @@ jobs:
       id-token: write
 
     steps:
+    - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


amoodie (approved):
Looks good to me, thanks for getting it set up with pypi. I'll delete the username and password secrets from the account once it is merged

PR #213, opened 2025-07-23
Reconcile shorelines

amoodie:
# PR to reconcile the API and naming around shoreline metrics.

This PR fixes the naming of several metrics related to shoreline characteristics, improves documentation about those metrics, and implements three new metrics for shoreline roughness.  

**This is a breaking change** from previous versions of DeltaMetrics / sandplover, because the function names have changed. 

~Depends on #132~ :heavy_check_mark: 

## changes to compute_shoreline_length

In this function, two things were changed. 

1. The input kwarg `origin` was deprecated and replaced with `start`; the old argument still works and the behavior of how it is handled has *not* changed. 
2. The new input kwarg `start` is expected to be in data dimensions (not indices). This is explained in the documentation.

## changes to compute_shoreline_roughness

This function was renamed to compute_shoreline_roughness_area. 

## new metrics for shoreline roughness
Multiple functions for computing shoreline roughness in subtly different ways were added. Below is an image from a documentation example that shows a comparison between these methods. 
image

## affecting multiple functions

- previously an error would rise if no shoreline / land area was determined that would leads to a divide by zero issue.
- now, these functions avoid the divide by zero, issue a warning, and return np.nan
- afftects compute_shoreline_distance, compute_shoreline_length, compute_shoreline_radius, compute_shoreline_roughness_variation, compute_shoreline_roughness_area, compute_shoreline_roughness_radius, compute_shoreline_roughness_OAM, and any functions that call these functions.

Todo
- [x] switch the naming between roughness and rugosity to be more consistent with engineering definitions of these terms.
- [x] add the roughness measure `OAM_45/OAM_120` from https://doi.org/10.1029/2022GL100355
- [x] add the roughness measure `count shoreline / mean_rad` from https://doi.org/10.5194/esurf-3-67-2015, 2015
- [x] fix the ordering of `origin` in `compute_shoreline_distance` to be consistent with everywhere else `(dim1, dim2)`
- [x] fix the `origin` kwarg for all of the functions to expect origin in data coordinates (not array indices)
- [x] change all to return nan and warning, when there are no pixels found. instead of erroring.
- [x] add tests for new functions
- [x] check all documentation, including an example document that compares all of the calculations


sandplover/plan.py
```diff
@@ -1270,84 +1326,174 @@ def compute_shoreline_roughness(shore_mask, land_mask, **kwargs):
...
+            category=UserWarning,
+            stacklevel=2,
+        )
 
     return rough
 
 
-def compute_shoreline_rugosity(shore_mask, **kwargs):
-    """Compute shoreline rugosity.
+def compute_shoreline_roughness_variation(shore_mask, origin=(0, 0)):
```
ericbarefoot: I thought we were deprecating origin? I still see it here, would you like to change this?

sandplover/plan.py
```diff
@@ -1364,23 +1510,339 @@ def compute_shoreline_rugosity(shore_mask, **kwargs):
...
+            category=UserWarning,
+            stacklevel=2,
+        )
+
+    return roughness
+
 
-    return rugosity
+def compute_shoreline_roughness_radius(
+    shore_mask, origin=(0, 0), calculate_length=False, **kwargs
```
ericbarefoot, line 1533: Also origin shows up here.

sandplover/plan.py
```diff
@@ -1270,84 +1326,174 @@ def compute_shoreline_roughness(shore_mask, land_mask, **kwargs):
...
+        _sm = shore_mask.values
+        _dx = float(
+            shore_mask[shore_mask.dims[0]][1] - shore_mask[shore_mask.dims[0]][0]
+        )
+    elif isinstance(shore_mask, np.ndarray):
+        _sm = shore_mask
         _dx = 1
     else:
-        raise TypeError(f"Invalid type {type(land_mask)}")
+        raise TypeError(f"Invalid type {type(shore_mask)}")
```
ericbarefoot, line 1378: This little block of code gets copy-pasted around a few times, should it be its own utility?

sandplover/plan.py
ericbarefoot: I also think "variation" makes sense here, but I do wonder if it implies dimensional variance, whereas you are computing dimensionless coefficient of variation.

One thing to consider would be renaming it `compute_shoreline_roughness_coef_of_var(...)`

ericbarefoot (approved):
Overall, I had no issues with merging this, the thing I focused on was thinking about the logic of the names. 

I think having the subset of names that separate out the different methods of computing roughness based on methodology seems to make a lot of sense.

I think it makes sense to deprecate the term rugosity in this and then also deprecate the original roughness calculation.

 I think I kind of understand the logic behind doing keyword arguments for the point of which you're using to measure all of the shoreline roughness metrics. I think what I'm confused about is why sometimes we use `start` and sometimes we use `origin`. It seems like that metrics based on area use `origin` whereas metrics based on length use `start`. Is this the case?

It might make some sense to keep both as an option and just allow them to be interchangeable aliases of each other? From my perspective, `start` doesn't make a whole lot of sense, but maybe I'm thinking about it from a purely delta perspective whereas I should be more flexible. Regardless I think it kind of makes sense to allow both. This is not a strong preference for both, but I do think that if we're not going allow both we should choose one and not have different keyword arguments for the same basic idea.

I think all of the function names make complete intuitive sense to me with the exception of `compute_shoreline_roughness_variation()` I get that it's supposed to be the coefficient of variation but the name just says variation which makes me immediately think of the variance rather than the non-dimensional coefficient of variation.

Happy to have a debate about the `start` vs `origin` question, but also happy for you to merge it, since I think that's a small part of the API we can gently change later or make more permissive as you have in the deprecation warnings.

sandplover/plan.py
amoodie: I changed the name of this function to `_coefvar`, thanks for that suggestion.

`origin` is deprecated as the input to `compute_shoreline_length`, because this function needs a `start` point, which need not be the origin of the delta (the apex). We still have both, but they have distinct meanings now, and in some cases both should be specified as different locations on the delta.

sandplover/plan.py
amoodie: probably, I will open an issue as a good first issue for someone to tackle.

amoodie:
Thanks for the review, @ericbarefoot.

- I renamed the function to `compute_shoreline_roughness_coefvar` per your suggestion. (I wanted to keep the suffix without additional underscores to match the others).
- I double checked the use of `origin` and `start`, and it is how I intended it. Basically, `origin` is used to refer to the point away from which rays radiate for some calculations, like the length of a delta `origin` is the delta apex. `start` is a distinct argument for the starting point of some algorithm. Previously, we used `origin` as the starting point for a line-tracing algorithm, but I found this to be confusing to use the same word for two distinct purposes. This PR makes it so (in `plan.py` at least) `origin` and `start` have a consistent meaning.

PR #218, opened 2025-08-12
order-agnostic dimensions validation (ignore leading 1-D vars) + derive cube shape from coords (fixes #216)

### LINKED ISSUE #216: Determining the shape of the DataCube currently relies on specific variable names
[Determining the shape of the DataCube currently relies on specific variable names.](https://github.com/sandpiper-toolchain/sandplover/blob/fe7cf387717abd14bd100a81e10cc627b0ce5389/sandplover/cube.py#L685C9-L696C40) Strangely, things seem to be working when the "x" "y" "time" names _don't_ match. This might be because another variable is found first that has a 3d shape and this is used, but I think it is possible a coordinate variable could come up first, and this would throw an error.

```
        # get shape from a variable that is not x, y, or time
        i = 0
        while i < len(self.variables):
            if (
                self.variables[i] == "x"
                or self.variables[i] == "y"
                or self.variables[i] == "time"
            ):
                i += 1
            else:
                _var = self.variables[i]
                i = len(self.variables)
```

To be data agnostic and support more of sandsuet standard, we should refactor this to work for a range of data sets. Tests are needed.

pthak006:
**Problem:** 
DataCube shape inference was effectively name/order-dependent. When dict inputs included a leading 1-D variable (e.g., station_id) ahead of a 3-D field, the loader validated user dimensions against that 1-D shape and crashed—even though a correct 3-D var was present.

**Diagnostic code for the problem:**

```
import numpy as np, sandplover as spl
dims = {'t_ax': np.arange(10), 'y_ax': np.arange(50), 'x_ax': np.arange(60)}
data = {'station_id': np.arange(100), 'temperature': np.random.rand(10,50,60)}
spl.cube.DataCube(data, dimensions=dims)

# raises: "Shape of `dimensions` at position 0 was 10, which does not match the variables dimensions (100,)."
```

**Fixes/Changes:**

1. DictionaryIO.get_known_coords

If any value is an xarray.DataArray, ignore dimensions and use the array’s dims/coords (preserves legacy behavior; fixes test_dimensions_ignored_if_xarray).

Otherwise, when dimensions are provided, validate against a 3-D reference variable (ndim==3), not the first dict entry. Keeps original error wording so tests stay green.

When dimensions are absent, infer dims from the first 3-D variable; error if none found.

**Follow-ups (possible)**
Consider a stricter validation of dimensions against all 3-D vars (not just the first reference).

2. Cube Classes Changes (related)

DataCube.__init__: set (_H,_L,_W) from coords (not from the first data variable).

BaseCube._read_meta_from_file: prefer dataio.dims/coords; avoid inferring from an arbitrary first var.

Minor: keys now returns list(self.dataset.keys()).

Local Run of pytest didn't result in any failed test cases.


amoodie:
:wave: hi @pthak006, thanks for submitting this PR and addressing an issue. 

I think the fix makes sense to me, but it would be great to see that proven with a couple of simple tests. Can you put together some new tests in `tests/cube_test.py` and `tests/io_test.py` that try to cover some of the cases that were an issue for #216 ?

For example, there are some tests for dictionaries [here](https://github.com/sandpiper-toolchain/sandplover/blob/5a7169f60842878e3b4e385b705ee6471112a4fd/tests/io_test.py#L157) that you could add to, for the cases where the 1D variable is first, and where no 3d variables are available at all. This will help you [bring up the coverage](https://github.com/sandpiper-toolchain/sandplover/pull/218#issuecomment-3181210053) as well.

But again, this looks to me like the fix would work, and thanks!

pthak006:
Thankyou @amoodie for the comments. Sure I'll try to add some new test cases in the provided test files.

pthak006:
> 👋 hi @pthak006, thanks for submitting this PR and addressing an issue.
> 
> I think the fix makes sense to me, but it would be great to see that proven with a couple of simple tests. Can you put together some new tests in `tests/cube_test.py` and `tests/io_test.py` that try to cover some of the cases that were an issue for #216 ?
> 
> For example, there are some tests for dictionaries [here](https://github.com/sandpiper-toolchain/sandplover/blob/5a7169f60842878e3b4e385b705ee6471112a4fd/tests/io_test.py#L157) that you could add to, for the cases where the 1D variable is first, and where no 3d variables are available at all. This will help you [bring up the coverage](https://github.com/sandpiper-toolchain/sandplover/pull/218#issuecomment-3181210053) as well.
> 
> But again, this looks to me like the fix would work, and thanks!


Hello @amoodie as you suggested I added some test cases in io_test.py and cube_test.py. Could you please run code coverage report once again ?

pthak006:
Hi @amoodie I think now the code coverage/ path coverage is above the threshold. If it looks good to you, we can merge the PR. Otherwise please let me know if I need to change some other things too.

amoodie:
Thanks @pthak006 for the PR!

Issue #43, opened 2021-03-17
Geometric mask `angular` does not work for grids with an odd number of cells in the `L` direction


amoodie:
Geometric mask `angular` does not work for grids with an odd number of cells in the `L` direction. I believe this is due to the slicing and creating of the meshgrid here:

https://github.com/DeltaRCM/DeltaMetrics/blob/fbfabb2b06f776625d5e5ad42cc563a194884b1b/deltametrics/mask.py#L1027

PR #221, opened 2025-10-07
Possible solution to Issue #43  Geometric mask 'angular' doesn't work for Odd L value 

pthak006:
1. GeometricMask.angular() could raise a broadcast error or produce a miscentered mask when the grid length L is odd (and in some cases when W > 2L). This PR makes the helper grid even-width and at least as wide as the requested crop so the center index is integral and the final crop always has shape (L, W).

**The main problem**: 

- For odd L, the helper x-extent has B = 2L + 1 columns (odd).

- The crop uses int(B/2 ± W/2), which floors a half-index center and can return a slice whose width ≠ W.

- With W > 2L, the helper field is narrower than the crop request, so slicing is clamped and the result also has width ≠ W.

- Final multiply self._mask *= anglemap then fails with a shape/broadcast error
- 

**Minimal problem reproduction code**:

```
import numpy as np
from sandplover.mask import GeometricMask
L, W = 51, 120   # odd L
gm = GeometricMask(np.ones((L, W)))
gm.angular(0, np.pi/3)  # before: ValueError broadcasting (L,W) with (L, <wrong>)
```

**Possible Solution Done**:

- Build a helper grid with width B = max(W, 2L), then force B even.

- Compute x = arange(-B//2, B//2), y = arange(0, W), evaluate angles, and crop a centered window of exactly W columns using integer math:

- left = B//2 - W//2

- right = left + W

- Multiply into self._mask with a guaranteed (L, W) boolean anglemap.

The overall behavior should remain same for L = even and for L = odd there can be up to a one-column difference at the far lateral edge due to removal of the half-index rounding artifact. 

Can add additional test code for the changes if needed.

### DIFF
--- a/sandplover/mask.py
+++ b/sandplover/mask.py
@@ -2277,11 +2277,6 @@ class GeometricMask(BaseMask):
 
         .. note::
 
-           Requires a domain with a width greater than 2x its length right now.
-           Function should be re-factored to be more flexible.
-
-        .. note::
-
            Currently origin point is fixed, function should be extended to
            allow for an input origin point from which the angular bounds are
            determined.
--- a/sandplover/mask.py
+++ b/sandplover/mask.py
@@ -2324,27 +2324,30 @@ class GeometricMask(BaseMask):
             >>> _ = ax[1].set_yticks([])
             >>> _ = ax[1].set_title("Mask * Topography")
         """
-        # width must exceed 2x length
-        if (self._L / self._W) > 0.5:
-            raise ValueError("Width of input array must exceed 2x length.")
+        # Helper grid notes:
+        # - We build a polar-angle field θ(y,x) on a rectangular helper grid, then
+        #   crop a centered (L, W) window from it.
+        # - Columns (x) span an EVEN number B so that the grid has a well-defined
+        #   integer center column; this makes the centered crop stable for any parity
+        #   of L and W.
+        # - Rows (y) follow L (not W) so the cropped window always has L rows, which
+        #   prevents broadcast errors when L > W.
 
-        # helper grid: at least W columns, and even so center is integral
         B = max(self._W, 2 * self._L)
         if B % 2:
             B += 1
 
-        # build helper coordinates
-        y = np.arange(self._W)[:, None]          # (W, 1)
-        x = np.arange(-B // 2, B // 2)[None, :]   # (1, B), total columns = B (even)
+        y = np.arange(self._L)[:, None]           # (L, 1)   row index (distance “up” from origin)
+        x = np.arange(-B // 2, B // 2)[None, :]   # (1, B)   centered symmetric columns
 
         theta = np.arctan2(x, y) - theta1 + np.pi / 2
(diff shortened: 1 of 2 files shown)


sandplover/mask.py
```diff
@@ -2324,20 +2324,29 @@ def angular(self, theta1=0, theta2=np.pi):
             >>> _ = ax[1].set_yticks([])
             >>> _ = ax[1].set_title("Mask * Topography")
         """
+        # width must exceed 2x length
         if (self._L / self._W) > 0.5:
             raise ValueError("Width of input array must exceed 2x length.")
```
amoodie: Can we get rid of this check given your improvements? I think this was a restriction we imposed to make things work for our use cases, but this assumption of domain shape obviously doesn't generalize.

amoodie (changes requested):
I think I understand, and it makes sense. In general, some brief explanation of what the helper coords are doing could be useful. 

Also, can you create a test for this new case ([added to the end of the GeometricMask tests](https://github.com/sandpiper-toolchain/sandplover/blob/7a06ee712fb1214e8ade6a430ae7933ef41330fc/tests/mask_test.py#L1644))? A test passing with a domain of odd number should be sufficient. If you also [address the check here](https://github.com/sandpiper-toolchain/sandplover/blob/7a06ee712fb1214e8ade6a430ae7933ef41330fc/sandplover/mask.py#L2327-L2328) and we can drop it, then a test showing a passing result for a domain with non-2x width would be sufficient.

sandplover/mask.py
ericbarefoot: I think so.

ericbarefoot (changes requested):
I agree with Andrew, and echo his request for a test.

pthak006:
Thanks a lot @amoodie @ericbarefoot  for the comments. As requested by you both I:

- Added some explanation comments for the helper grid
- Removed the check the condition that was requested for. However if we do that we silently assume **rows = W** (because of `y = np.arange(self._W)[:, None]`). That assumption was safe only because the old guard forced “W big enough". Once we drop the guard, cases with **L > W** become legal—and then the math breaks.

One example which will break because of this:
```
L, W = 60, 40        # now valid once the guard is gone
arr = np.zeros((L, W))
gmsk = GeometricMask(arr)
gmsk.angular(0, np.pi/2)   # anglemask has 40 rows, mask expects 60 → broadcast error
```
- To cover for this we need to do just one change in the current code:

`y = np.arange(self._L)[:, None]           # rows follow L, not W`

- I removed an old test case which checked for bad dims previously and added a new test case that checks for odd length case.

amoodie (approved):
Looks great!! Thanks!!

One small change: remove the note at lines 2278--2281, as this will be outdated after merge.

I'll merge as soon as that's done, thanks!

pthak006:
Thanks @amoodie for the reply. I did remove the comment lines you talked about. Sorry I forgot to remove it when I did the changes.

amoodie:
🎉

PR #222, opened 2025-11-13
Possible fix to issue #3: memory check before loading large variable into memory

pthak006:
### Problem
Users could accidentally load 50+ GB datasets into memory without warning by calling `cube.read(True)` or `cube.read('large_variable')`, potentially causing system crashes or severe performance degradation.

### Solution
Implemented memory safety checks that prevent loading variables exceeding 80% of available RAM, with an override option for intentional large loads.

#### 1. Memory Safety Implementation
- **`NetCDFIO.read(var, force=False)`** - Added memory checking logic before loading data:
  - Calculates variable size from xarray metadata (dtype × shape)
  - Compares against 80% of currently available system memory (via `psutil`)
  - Issues a clear warning and **does not load** if threshold exceeded
  - Provides guidance on using `force=True` to override
  
- **`BaseCube.read(variables, force=False)`** - Updated to accept and pass through `force` parameter to support memory-safe loading of single or multiple variables

#### 2. Dead Code Cleanup
- Removed unused `read` parameter from:
  - `BaseCube.__init__()`
  - `DataCube.__init__()`
  - `StratigraphyCube.__init__()`
  
This parameter was documented but never actually used during initialization. Removed it to eliminate confusion and clarified in docstrings that data is never loaded during cube instantiation—users must explicitly call `.read()`.


sandplover/cube.py
ericbarefoot, line 1: I think I'm not quite sure why we would purge the `read()`  invocation from `__init__()`. I think I see that `read()` is actually called in line 888, and then it gets `super()` called by inheriting classes on init. So It seems like it does in fact load on init. 

But I think that's kind of a different question, because I don't really have an issue with requiring `read()` to happen explicitly. @amoodie what do you think?

ericbarefoot (commented):
Main question is whether we should make `read()` an explicit action. I think that's generally not a bad idea, but it could be convenient to have it read on `__init__()`. I will defer to @amoodie on whether to add this additional step for the user.

pthak006:
Hi @ericbarefoot @amoodie I looked into the issue of read() parameter that I removed and also looked into the whole IO module. Also looked into the memory safe loading. Here are the things I found and changes I have made:

- I reverted and introduced back the read() parameters in different Cube classes that I removed. I actually found that the read parameter was intended for situations where you instantiate a cube variable and immediately want to load some/all variables into the memory. eg:

    ```
    # Just connect, don't load data (lazy)
    cube = DataCube('/path/to/file.nc')
    
    # Connect AND immediately load all variables into memory
    cube = DataCube('/path/to/file.nc', read=True)
    
    # Connect AND immediately load specific variables
    cube = DataCube('/path/to/file.nc', read=['eta', 'velocity'])
    ```
However it was never used. To make sure it works as intended I made changes [here](https://github.com/sandpiper-toolchain/sandplover/pull/222/files#diff-72fa09536438ac4b7cdfa857f9f266354de306342e41f42a5bcf4ca53fd2dc0dR98)

- Second change I made was [here](https://github.com/sandpiper-toolchain/sandplover/pull/222/files#diff-72fa09536438ac4b7cdfa857f9f266354de306342e41f42a5bcf4ca53fd2dc0dR766). The reason is that previously DataCube's __getitem__ method was directly invoking xarray's variable loading method without checking the _in_memory_data cache. NetCDFIO's getitem method is designed for that but the DataCube never utilized that. 

- The memory safety code I wrote in io.py are still unchanged. I think with these changes the cube.read() pathway is memory safe. DictionaryIO is used when the data is already in memory, BaseIO and FileIO are abstract classes and hence memory check is not needed in these IO classes. 

- However it is important that this memory safety only works when someone is invoking the read pathway that is invoking functions like following:

```
cube.read('var')
cube.read(True)
DataCube('/path.nc', read=True)
```

In case someone does invoke xarray load directly like the following cases there's no memory protection then:

```
cube['var'].data
cube['var'].values
cube['var'].load()
np.array(cube['var'])
cube.export_frozen_variable('var')
```

I can see in the doc strings of cube.py classes that a CubeVariable class was intended to be used for this purpose which will intercept .data, .values, and .load() calls but was never implemented. I didn't introduce it into this PR's code as it seems to be outside of this PR's scope.

ericbarefoot (approved):
This looks good to me. Let's merge ahead. Thanks @pthak006 !

pthak006:
No worries Professor, let me know when I can resume work on the same
starting in Fall. I will also keep an eye out on the codebase and try to
actually see if some minor issues I can resolve easily and pick them up if
I have time.

-Partha

PR #228, opened 2026-01-30
Fix tests

amoodie:
this PR will address failing builds on Github Actions

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -36,11 +36,11 @@ jobs:
       uses: actions/setup-python@v5
       with:
         python-version: ${{ matrix.python-version }}
-    - name: Install dependencies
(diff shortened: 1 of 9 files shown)


amoodie:
I believe the issue is a conflict between netcdf4 and h5py libraries. 
https://github.com/h5py/h5py/issues/2453

Will try pinning one or both to earlier versions.

amoodie:
This PR fixes SOME of the issues with our failing builds. 

There remain issues with "This probably means that Tcl wasn't installed properly." which seems to be a widespread and unexplained error on Github actions. 

There remain issues with MacOS builds getting rejected from downloading by the Figshare API, which seems to be a widespread error on Github actions.

Neither of these are issues with our build, so I am going to merge the (small) improvements in this PR.

PR #250, opened 2026-07-24
change `meta` to `aux` and allow it to be specified during instantiation

amoodie:
Rename `meta` to `aux` at the `Cube` and `IO` levels. Enable `auxdata` to be specified when instantiating a `DataCube` and this argument is passed to the `NetCDFIO` file handler to set up this subgroup as the `cube.aux` attribute. 

No warning is issued if nothing is passed and nothing is detected for auxdata.

Note: this PR does keep automatic detection of a few subgroup names for now, just for backwards compatibility and a smoother transition, but warns that this will be removed in the future. 

todo
- [ ] implement this auxdata argument for `DictionaryIO`
- [x] rename cubes `_read_meta_from_file` to something more intuitive
- [x] clean up IO layer with making it more clear what is private

### DIFF
--- a/sandplover/cube.py
+++ b/sandplover/cube.py
@@ -41,7 +41,7 @@ class BaseCube(abc.ABC):
 
     """
 
-    def __init__(self, data, read=(), varset=None, dimensions=None):
+    def __init__(self, data, auxdata=None, read=(), varset=None, dimensions=None):
         """Initialize the BaseCube.
 
         Parameters
@@ -52,6 +52,10 @@ class BaseCube(abc.ABC):
             output from the pyDeltaRCM model. Alternatively, pass a
             :obj:`dict` with keys indicating variable names, and values with
             corresponding t-x-y `ndarray` of data.
+
+        auxdata : :obj:`str`, optional
+            The information `data` is searched for a key matching the string
+            `auxdata`, and if found, this key is assigned to `cube.aux`.
 
         read : :obj:`bool`, optional
             Which variables to read from dataset into memory. Special option
@@ -69,7 +73,7 @@ class BaseCube(abc.ABC):
         if type(data) is str:
             # handle a path to netCDF file
             self._data_path = data
-            self._connect_to_file(data_path=data)
+            self._connect_to_file(data_path=data, auxdata_path=auxdata)
             self._read_meta_from_file()
         elif type(data) is dict:
             # handle a dict, arrays set up already, make an io class to wrap it
@@ -109,7 +113,7 @@ class BaseCube(abc.ABC):
         """
         ...
 
-    def _connect_to_file(self, data_path):
+    def _connect_to_file(self, data_path, auxdata_path):
         """Connect to file.
 
         This method is used internally to send the ``data_path`` to the
@@ -117,9 +121,9 @@ class BaseCube(abc.ABC):
         """
         _, ext = os.path.splitext(data_path)
         if ext == ".nc":
-            self._dataio = NetCDFIO(data_path, "netcdf")
+            self._dataio = NetCDFIO(data_path, auxdata_path, "netcdf")
         elif ext == ".hdf5":
-            self._dataio = NetCDFIO(data_path, "hdf5")
+            self._dataio = NetCDFIO(data_path, auxdata_path, "hdf5")
         else:
             raise ValueError('Invalid file extension for "data_path": %s' % data_path)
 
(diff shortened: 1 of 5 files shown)


sandplover/io.py
```diff
@@ -259,25 +289,22 @@ def connect(self):
...
-                UserWarning,
-                stacklevel=2,
-            )
-        else:
-            warnings.warn(
-                "No associated metadata was found in the given data file.",
-                UserWarning,
-                stacklevel=2,
-            )
-            self.meta = None
```
ericbarefoot, line 280: Do you want to consider trapping these still in case someone manually assigns them? The group could get redirected to auxdata and then a deprecation warning raised.

sandplover/section.py
```diff
@@ -1684,8 +1684,8 @@ def _compute_section_coords(self):
                         5,
                     )
                 else:
-                    # extract L0 from the cube metadata
-                    land_width = int(self._underlying.meta["L0"])
+                    # extract L0 from the cube aux data
+                    land_width = int(self._underlying.aux["L0"])
```
ericbarefoot, line 1688: Claude helpfully points out that there's an edge case where someone has an aux field without a key `"L0"`, what then? This raises with an unhelpful KeyError.

sandplover/section.py
```diff
@@ -1934,8 +1934,8 @@ def _compute_section_coords(self):
...
```
ericbarefoot, line 1938: Same idea here.

sandplover/cube.py
```diff
@@ -69,18 +73,20 @@ def __init__(self, data, read=(), varset=None, dimensions=None):
...
+            self._dataio = DictionaryIO(
+                data, dimensions=dimensions, auxdata_path=auxdata
+            )
+            self._read_coords_dims_variables_from_dataio()
         elif isinstance(data, DataCube):
             # handle initializing one cube type from another
             self._data_path = data.data_path
             self._dataio = data._dataio
-            self._read_meta_from_file()
+            self._read_coords_dims_variables_from_dataio()
```
ericbarefoot, line 89: Claude points out that this creates a view, not a copy, so anything that you change about aux on the returned dataset will also alter the origin dataset. Is that the intention? I think one could make an argument either way.

sandplover/cube.py
ericbarefoot, line 201: In this read function, a list will actually error out because it gets caught in the else at the end, and you never get to the for var in variables loop.

sandplover/cube.py
```diff
@@ -109,22 +115,8 @@ def __getitem__(self, var):
         """
         ...
 
-    def _connect_to_file(self, data_path):
```
ericbarefoot, line 112: I think that I understand that this backend is removed in favor of always using netcdfio

ericbarefoot (changes requested):
I checked for myself and ctrl-f ed around to make sure there were no more instances of `meta` in the code. 

I like the usage improvement very much to be able to specify aux as whatever you want. 

I decided to do an AI review for fun, and it turned up a few things in the diff, but also pointed out that several things changed here expose other issues. One of those was in 

https://github.com/sandpiper-toolchain/sandplover/blob/34ddac0f13b51a7d4672cfeaa51b2f8edc95028d/sandplover/utils.py#L68

where we pass obj to AttributeError as a positional argument, but that's not the usage for that. instead should be ` __init__(message, obj=obj)`

Otherwise, I am glad you did this!

sandplover/io.py
amoodie: I did it this way at first: automatically detecting any of the common groups we have used in the past, assigning that to `.aux` and raising a deprecation warning. 

It worked fine that way, but it also raised warnings all over our testing suite. I decided to it would be easier to just have this as a breaking change now (still in alpha). If you think we should raise a warning, I can add that back and change all of the tests to expect warnings.

sandplover/section.py
amoodie: yes, but this is a different issue we also plan to address (remove automatic detection of things), rather than changing the name of the meta field. I'm trying to break this into logical steps of change.

sandplover/cube.py
amoodie: I just moved this logic to the IO layer, where it belongs

sandplover/cube.py
amoodie: It will alter the instance of the opened dataset, but not alter the dataset on file, nor does it affect other cubes accessing the same dataset on file. But, a change would affect a derived StratigraphyCube, because it shares the same IO instance with the parent DataCube. 

I'm not sure it matters to use a view or copy for that reason --- we don't support the concept of having multiple IO layers for a singe cube. In either case, it isn't really related to this PR, which just changes the name of the method to be more clear about what it actually does.

sandplover/cube.py
amoodie: yeah I think you are correct, will open an issue for that bug to be addressed in another PR

amoodie:
Thanks for the review @ericbarefoot! 

I think most of the issues identified are not related to the changes here though (even though some relevant lines may have been touched). The only remaining question I have, I think, is whether we want to make the breaking change over a few minor versions (your comment about raising warnings, which @mcflugen also asked about on the call yesterday), or just make the breaking change with v0.6.0.

amoodie:
Hearing no objections, I am going to merge this breaking change for version 0.6.0

PR #251, opened 2026-07-30
compensation returns improvement / fix

amoodie:
revise private compensation calculation to be more consistent with expected returns.


ericbarefoot (approved):
Yeah this looks good to me!

